PR #28, opened 2014-12-18
Fixes #8756: Generates consumer cert RPM on the Capsule.

ehelms:


stbenjam:
It doesn't look like this explicitly requires httpd package - what brings that in, pulp?  

I think we should maybe add `ensure_packages(['httpd'])` to `certs::katello` - thoughts?

manifests/init.pp
```diff
@@ -342,4 +341,9 @@
     }
 
   }
+
+  class { 'certs::katello':
+    deployment_url  => $capsule::rhsm_url,
+    rhsm_port       => $capsule::rhsm_port
```
stbenjam: Defined in params.pp, but missing them in the actual capsule class

manifests/init.pp
```diff
@@ -11,7 +11,7 @@
 #
 # $pulp_master::                    whether the capsule should be identified as a pulp master server
 #
-# $pulp_admin_password::            password for the Pulp admin user. It should be left blank so that a random password is generated
+# $pulp_admin_password::            passowrd for the Pulp admin user.It should be left blank so that random password is generated
```
stbenjam: ?

manifests/init.pp
```diff
@@ -266,7 +265,7 @@
     }
   }
 
-  $foreman_proxy = $tftp or $dhcp or $dns or $puppet or $puppetca or $realm or $pulp
+  $foreman_proxy = $tftp or $dhcp or $dns or $puppet or $puppetca or $realm
```
stbenjam: This and the change above makes me think you need to rebase your checkout?

ehelms:
Updated. Will require https://github.com/Katello/puppet-certs/pull/43

manifests/init.pp
```diff
@@ -11,7 +11,7 @@
 #
 # $pulp_master::                    whether the capsule should be identified as a pulp master server
 #
-# $pulp_admin_password::            password for the Pulp admin user. It should be left blank so that a random password is generated
+# $pulp_admin_password::            password for the Pulp admin user.It should be left blank so that random password is generated
```
stbenjam: Looks like this one needs rebasing too, missing b23ca86

manifests/init.pp
```diff
@@ -252,7 +258,6 @@
     } ~>
     class { 'puppet':
       server                      => true,
-      server_ca                   => $puppetca,
```
stbenjam: Why?

stbenjam:
Tested with the other 2 PR's and it seems to work fine :+1: , just a couple questions here and some minor puppet-lint failures in Travis.

ehelms:
@stbenjam Up0dated

manifests/init.pp
```diff
@@ -11,7 +11,7 @@
 #
 # $pulp_master::                    whether the capsule should be identified as a pulp master server
 #
-# $pulp_admin_password::            password for the Pulp admin user. It should be left blank so that a random password is generated
+# $pulp_admin_password::            password for the Pulp admin user. It should be left blank so that random password is generated
```
stbenjam: Nitpicky, but why does your git client hate this line? The original is better

stbenjam:
ACK pending fixing the $pulp_admin_password description

PR #29, opened 2015-01-13
fixes #8909 - correct version numbers of dependent modules

stbenjam:
Librarian won't update katello-installer's capsule module because metadata.json are referencing versions which don't exist, for example, the latest theforeman-dns is 1.4.  Not sure how it worked before?

Anyway, we don't need them here, Librarian should pull in the dependencies based on `theforeman-foreman_proxy`.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -32,26 +32,6 @@
        "version_requirement": ">= 1.5.0"
     },
     {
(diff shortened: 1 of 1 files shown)


ehelms:
ACK

PR #30, opened 2015-01-21
Fixes #7745: Deploy reverse proxy for RHSM traffic.

ehelms:


ehelms:
Requires - https://github.com/Katello/puppet-certs/pull/45

ehelms:
Also requires -- https://github.com/Katello/katello/pull/4949

manifests/init.pp
```diff
@@ -32,6 +32,10 @@
 #                                   Setting this to anything non-empty causes
 #                                   the apache vhost to set up a proxy for all
 #                                   certificates pointing to the value.
+# $parent_reverse_proxy::           Add reverse proxy to the parent
```
stbenjam: Can this just be $reverse_proxy? Since this will be exposed as Kafo parameters, i like `--capsule-reverse-proxy=true` instead of `--capsule-parent-reverse-proxy=true`.  It also needs to match the name of the feature being enabled, if we will some day have some automagic Smart Proxy + Foreman plugins...

Speaking of those, what are these? @dustints?

https://github.com/Katello/foreman_reverse_proxy
https://github.com/Katello/smart-proxy-http-reverse-proxy-plugin
https://github.com/Katello/smart-proxy-templates-plugin
https://github.com/Katello/smart-proxy-rhsm-plugin

smart-proxy-http-reverse-proxy-plugin looks the most active, but I don't follow the name.  Foreman's Smart Proxy plugin names correlate directly to the Feature name.  This should be `smart_proxy_reverse_proxy` so the feature becomes "Reverse Proxy" in the Foreman UI.

manifests/init.pp
```diff
@@ -213,6 +223,18 @@
     validate_present($foreman_oauth_secret)
   }
 
+  if $pulp or $capsule::parent_reverse_proxy {
```
stbenjam: $capsule not needed

manifests/init.pp
```diff
@@ -213,6 +223,18 @@
...
+    class { 'certs::apache':
+      hostname => $capsule_fqdn
+    } ~>
+    Class['certs::capsule'] ~>
+    class { 'capsule::reverse_proxy':
+      path => '/',
+      url  => "${foreman_url}/",
+      port => $capsule::parent_reverse_proxy_port
```
stbenjam: Ditto

manifests/reverse_proxy.pp
```diff
@@ -0,0 +1,50 @@
+#Adds http reverse-proxy to parent conf
+class capsule::reverse_proxy (
+
+  $path = '/',
+  $url  = "https://${capsule::parent_fqdn}/",
+  $port = $capsule::params::parent_reverse_proxy_port
```
stbenjam: This will always pull in the default value, not the value in `capsule`. 

If this class will always be wrapped by `capsule`, I wouldn't even parameterize, especially since you're "building" a URL in the class definition, I think that doesn't belong there.

manifests/init.pp
dustints: @stbenjam made some effort to integrate templating with the reverse proxy. I'll close those PRs for now.

ehelms:
Updated.

manifests/init.pp
```diff
@@ -200,17 +222,78 @@
...
+  class { 'capsule::install': } ~>
+  class { 'certs::foreman_proxy':
+    hostname => $capsule_fqdn,
+    require  => Package['foreman-proxy'],
+    before   => Service['foreman-proxy'],
+  } ~>
+  class { 'certs::katello':
+    deployment_url => $capsule::rhsm_url,
+    rhsm_port      => $capsule::rhsm_port
   }
```
stbenjam: This conflicts with puppet-katello:

```
[ERROR 2015-01-26 19:06:05 main]  Duplicate declaration: Class[Certs::Katello] is already declared in file /vagrant/katello-installer/modules/katello/manifests/init.pp:63; cannot redeclare at /vagrant/katello-installer/modules/capsule/manifests/init.pp:234 on node katello-centos7.example.com
```

stbenjam: Disregard, it was removed - needed to update that module in my katello-installer

stbenjam:
Tested! Works!

ACK

PR #31, opened 2015-01-23
fixes #8991 - configure templates proxy 

stbenjam:
Requires ~~https://github.com/Katello/puppet-capsule/pull/26/files~~ and ~~https://github.com/Katello/puppet-certs/pull/41~~

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -19,7 +19,13 @@
 #
 # $pulp_oauth_secret::              OAuth secret to be used for Pulp REST interaction
 #
-# $foreman_proxy_port::             Port on which will foreman proxy listen
+# $foreman_proxy_port::             SSL port on which foreman proxy will listen
+#                                   type:integer
+#
+# $foreman_proxy_http::             Foreman proxy listen on HTTP
+#                                   type:boolean
+#
+# $foreman_proxy_http_port::        HTTP port on which foreman proxy will listen
 #                                   type:integer
 #
 # $puppet::                         Use puppet
@@ -127,6 +133,9 @@
 #
 # $rhsm_url::                       The URL that the RHSM API is rooted at
 #
+# $templates::                      Enable templates proxying feature
+#                                   type:boolean
+#
 class capsule (
   $parent_fqdn                   = $capsule::params::parent_fqdn,
   $certs_tar                     = $capsule::params::certs_tar,
@@ -138,6 +147,8 @@ class capsule (
   $pulp_oauth_secret             = $capsule::params::pulp_oauth_secret,
 
   $foreman_proxy_port            = $capsule::params::foreman_proxy_port,
+  $foreman_proxy_http            = $capsule::params::foreman_proxy_http,
+  $foreman_proxy_http_port       = $capsule::params::foreman_proxy_http_port,
 
   $puppet                        = $capsule::params::puppet,
   $puppetca                      = $capsule::params::puppetca,
@@ -192,6 +203,7 @@ class capsule (
 
   $rhsm_url                      = $capsule::params::rhsm_url,
 
+  $templates                     = $capsule::params::templates,
   ) inherits capsule::params {
 
   validate_present($capsule::parent_fqdn)
@@ -215,10 +227,16 @@ class capsule (
 
   if $pulp_master or $pulp {
     foreman_proxy::settings_file { 'pulp':
-      template_path  => 'capsule/pulp.yml'
+      enabled       => $pulp_master,
+      listen_on     => 'https',
+      template_path => 'capsule/pulp.yml',
+
     }
+
     foreman_proxy::settings_file { 'pulpnode':
-      template_path => 'capsule/pulpnode.yml'
+      enabled       => $pulp,
+      listen_on     => 'https',
+      template_path => 'capsule/pulpnode.yml',
     }
   }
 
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -280,6 +285,14 @@
...
       before   => Service['foreman-proxy'],
+    } ~>
+
+    # Deploy client certificates for foreman proxy to talk to master
+    class { 'certs::foreman':
+      configure_foreman => false,
+      owner             => 'foreman-proxy',
+      client_cert       => $foreman_client_cert,
+      client_key        => $foreman_client_key,
     }
```
stbenjam: Actually this probably blows up in the case when installing the dependent capsule, will have to think of another way to do it, not sure how exactly :-\

I want to use RPM's already in the certs-tar on existing capsules.

manifests/init.pp
```diff
@@ -325,9 +329,11 @@
       register_in_foreman   => $register_in_foreman,
       foreman_base_url      => $foreman_url,
       registered_proxy_url  => "https://${capsule_fqdn}:${capsule::foreman_proxy_port}",
+      trusted_hosts         => undef,
```
iNecas: Why do we need to add this?

stbenjam: Because https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/params.pp#L38 and http://projects.theforeman.org/issues/9102

iNecas: Then it would be probably better to keep that in separate commit with the proper reference number

stbenjam: Sorry? This isn't fixing 9102, and 9102 isn't going to get fixed anytime soon.

This is *required* to use the templates plugin today, you need to disable trusted hosts...and thats what this line is doing.

stbenjam:
Ok, @iNecas I updated to include the setting to enable HTTP, so this now requires https://github.com/theforeman/puppet-foreman_proxy/pull/140.

When that is merged:

  * Templates module will listen by default on HTTP and HTTPS
  * Everything else only on HTTPS.  

Trusted hosts is disabled so arbitrary clients can connect and get templates, but note that SSL verification is still restricting access to other services.  As mentioned above, RM#9102 would be nice, but it's not required 
here.

manifests/init.pp
iNecas: You can use 'Refs #issue' instead of 'Fixes #issue' to have the PR tracked,but not marking the issue as fixed.
The advantage is also it might get into the master sooner than the rest of this changes (that I've just made you to be dependent on https://github.com/theforeman/puppet-foreman_proxy/pull/140 :) I guess other capsule features also need this workaround to start working properly again, right?

stbenjam: I still don't really follow, what are you asking me to do?

stbenjam: You want a separate PR for `trusted_hosts => undef` or something? It's not really related to #9102, it's a part of enabling the templates plugin.

Someday if 9102 is taken, we need a PR here to enable trusted hosts to be `[$::capsule_fqdn, $::parent_fqdn]` and enable some future `allow_all` to true on the templates, but that is not even on the horizon for 1.8.

iNecas: Ok, if it's needed just for this PR, just ignore my previous comments :)

stbenjam: Really, we should've had a PR to address this when https://github.com/theforeman/puppet-foreman_proxy/commit/952f6bd7c83470cdd25f38f6d768bc4bf10e6e95 was merged, but it doesn't matter now.

All the cats I'm herding now are going to end up in katello-installer in a single PR.

stbenjam:
Tests will pass when the corresponding `puppet-foreman_proxy` PR is merged.

manifests/init.pp
stbenjam: And sorry, updated because this should be `[]` instead of undef.

stbenjam:
theforeman/puppet-foreman_proxy#140 was merged.

@ehelms @iNecas Can one of you take a look? I'd like to get this and https://github.com/Katello/puppet-capsule/pull/30 merged today, so I can do end-to-end testing tomorrow of provisioning a client only through the capsule.

ehelms:
@stbenjam Needs a rebase after what I just merged, apologies :)

stbenjam:
Updated

manifests/params.pp
```diff
@@ -26,6 +26,10 @@
   $puppet_ca_proxy               = ''
 
   $foreman_proxy_port            = '9090'
+
+  $foreman_proxy_http            = true
+  $foreman_proxy_http_port       = '8000'
```
ehelms: Is there an selinux rule in place already for this port?

stbenjam: Yes - it's the foreman default.  I imagine it could be changed to be `$foreman_proxy::params::http_port`.

```
[root@mahcapsule ~]# setenforce 1
[root@mahcapsule ~]# service foreman-proxy restart
Stopping foreman-proxy:                                    [  OK  ]
Starting foreman-proxy:                                    [  OK  ]
[root@mahcapsule ~]# telnet localhost 8000
Trying ::1...
telnet: connect to address ::1: Connection refused
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
```

ehelms:
ACK

PR #33, opened 2015-02-03
Refs #8710 - install katello-debug by default

lzap:
Belongs to https://github.com/Katello/katello/pull/4970

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -7,4 +7,7 @@ class capsule::install {
     }
   }
 
+  package{ ['katello-debug']:
+    ensure => installed,
+  }
 }


lzap:
Hello, anyone? @iNecas @jlsherrill Thanks!

lzap:
Context: I've created foreman-debug and katello-debug subpackages. We want to install them on the Server as well as on the Capsules. Katello-debug has dependency of foreman-debug (it is technically an "extension" to it).

stbenjam:
Yea this is fine, but the package has to exist before we merge it, it's not in the nightly repos yet.

I imagine because the Pipeline is red? http://ci.theforeman.org/view/Katello%20Pipeline/

manifests/install.pp
```diff
@@ -5,6 +5,10 @@
     package{ ['rubygem-smart_proxy_pulp']:
       ensure => installed,
     }
+
+    package{ ['katello-debug']:
+      ensure => installed,
```
stbenjam: Actually why is this in the if block for pulp? We should always install `katello-debug`, or should we?

iNecas:
What @stbenjam sais

lzap:
Right fixed.

stbenjam:
Looks like the packages are there now, so ACK

stbenjam:
@lzap You can merge your own PR's, it's how it's done in the Katello GitHub org, but I'll merge this one

PR #34, opened 2015-02-04
refs #8175 - configure dispatch router on pulp and pulp nodes

stbenjam:
Requires ~~https://github.com/Katello/puppet-certs/pull/51~~ and ~~https://github.com/Katello/puppet-qpid/pull/15~~ and ~~https://github.com/Katello/puppet-qpid/pull/17~~

### DIFF
--- a/manifests/dispatch_router.pp
+++ b/manifests/dispatch_router.pp
@@ -1,0 +1,74 @@
+# == Class: capsule::dispatch_router
+#
+# Install and configure Qpid Dispatch Router
+#
+class capsule::dispatch_router (
+) {
+
+  class { 'qpid::router': }
+
+  # SSL Certificate Configuration
+  class { 'certs::qpid_router':
+    require => Class['qpid::router::install'],
+  } ~>
+  qpid::router::ssl_profile { 'client':
+    ca   => $certs::ca_cert,
+    cert => $certs::qpid_router::client_cert,
+    key  => $certs::qpid_router::client_key,
+  } ~>
+  qpid::router::ssl_profile { 'server':
+    ca   => $certs::ca_cert,
+    cert => $certs::qpid_router::server_cert,
+    key  => $certs::qpid_router::server_key,
+  }
+
+  # Listen for katello-agent clients
+  qpid::router::listener { 'clients':
+    addr        => $capsule::qpid_router_agent_addr,
+    port        => $capsule::qpid_router_agent_port,
+    ssl_profile => 'server',
+  }
+
+  # Act as hub if pulp master, otherwise connect to hub
+  if $capsule::pulp_master {
+    qpid::router::listener {'hub':
+      addr        => $capsule::qpid_router_hub_addr,
+      port        => $capsule::qpid_router_hub_port,
+      role        => 'inter-router',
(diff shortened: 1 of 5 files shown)


manifests/dispatch_router.pp
```diff
@@ -0,0 +1,52 @@
...
+#
+class capsule::dispatch_router (
+) {
+
+  class { 'qpid::router': }
+
+  qpid::router::ssl_profile { 'router-ssl':
+    ca   => $certs::ca_cert,
+    cert => $certs::params::messaging_client_cert,
+    key  => $certs::params::messaging_client_cert,
```
ehelms: Avoid referencing params.

manifests/dispatch_router.pp
```diff
@@ -0,0 +1,52 @@
...
+    qpid::router::connector { 'hub':
+      addr        => $capsule::parent_fqdn,
+      port        => $capsule::qpid_router_hub_port,
+      ssl_profile => 'router-ssl',
+      role        => 'inter-router',
+    }
+  }
+
+  # Listen for katello-agent clients
+  qpid::router::listener { 'clients':
```
ehelms: Maybe move this above the if-else so thats it not lost at the bottom given the size of the if/else ?

manifests/init.pp
```diff
@@ -136,6 +136,21 @@
 # $templates::                      Enable templates proxying feature
 #                                   type:boolean
 #
+# $qpid_router::                    Configure qpid dispatch router
```
ehelms: This parameter doesn't look like it is used anywhere.

manifests/dispatch_router.pp
stbenjam: I guess I need some advice here.  For the qpid dispatcher, I need 2 ssl profiles on the hub - one for the server and one for the connection to the existing broker.

I'm not really sure what to use, I don't want to create yet more certificates, can I use the qpid-broker ones?

manifests/init.pp
stbenjam: It is now

stbenjam:
Note, tests will fail until the other PR's mentioned in the top comment.

manifests/dispatch_router.pp
```diff
@@ -0,0 +1,74 @@
...
+# Install and configure Qpid Dispatch Router
+#
+class capsule::dispatch_router (
+) {
+
+  class { 'qpid::router': }
+
+  # SSL Certificate Configuration
+  class { 'certs::qpid_router':
+    require => Package[$qpid::router::router_packages],
```
ehelms: Couldn't we just require `Class['qpid::config']` ? Should that class itself not be requiring this package to be present?

stbenjam: I don't think I can't anchor that to happen before because of puppet-concat's voodoo.

`Class['qpid::config']` contains the actual file entry for the concat bits, and the certs need to be done before the qdrouterd.conf is laid down.

stbenjam: I could require `Class['qpid::install']`, that might be cleaner

ehelms:
ACk

PR #37, opened 2015-03-09
Fixes #9157: Allow customizing the crane port.

ehelms:


ehelms:
Requires https://github.com/Katello/puppet-crane/pull/2

jlsherrill:
This port # seems to be hardcoded into katello:

jlsherrill:
We'd also need to somehow know the port on a per-capsule basis which is more complicated....

We could allow them to change it but it must be uniform across their capsules.  Or we could contact the capsules and fetch the port prior to deployment

jlsherrill:
Given all that it may not be worth doing?

ehelms:
You have me convinced.. until someone produces a viable use case for this I am all for not supporting it until then.

PR #41, opened 2015-04-02
Refs #9995: Add monit monitoring of the qdrouterd service

ehelms:


ehelms:
This is to account for the fact that under certain conditions or stresses the qdrouterd service can core dump and shut down. I debated the best place to put this as it is sort of a 'hack' or work around for an issue and decided here as it would be easy to remove and seems like a function of our use case less so than embedding it into the puppet-qpid module itself. Further, its worth noting the current '5 second' window for monit to check the status and restart qdrouterd - possibly this should be shorter? or could be longer?

manifests/dispatch_router.pp
```diff
@@ -7,6 +7,17 @@
 
   class { 'qpid::router': }
 
+  if $monitor_qpid_router {
```
stbenjam: I assume this should be in $capsule::monitor_qpid_router? Ditto for $monitor_interval

stbenjam:
Other than the minor errors from Travis, it looks good.

How serious is the issue? I haven't seen it myself and I'm not sure what exactly it takes to reproduce other than heavy usage over a few hours?  Do we have news from the qpid team about their efforts to uncover the cause?

I do prefer that we not host workarounds in Katello for other software's problems.  This opens up such a can of worms :-1: :disappointed: 

These capsule class parameters are going to end up in the Kafo answer file -- and we have no solution to deal with removing those.  We also have to clean up after ourselves at some point later and remove monit.

ehelms:
1) We could get around the parameters by not letting them configurable (but we have this issue anyway)

2) I'll let @mccun934 explain and give more details on the severity outside of what the BZ provides (https://bugzilla.redhat.com/show_bug.cgi?id=1205893). The qpid team is attempting to diagnose the root cause.

3) I don't know anyone that enjoys carrying work around for other software, but there is always the lesser of two evils between giving users broken bits or fixing them. Without a true root cause fix, we'd either have to release with this hole patch or cross our fingers and hope for the best if our users do run into this issue.

stbenjam:
> 1) We could get around the parameters by not letting them configurable (but we have this issue anyway)

I think that's easier to clean up later

> 2) I'll let @mccun934 explain and give more details on the severity outside of what the BZ provides (https://bugzilla.redhat.com/show_bug.cgi?id=1205893). The qpid team is attempting to diagnose the root cause.

I saw they got a core dump finally so do we have the luxury of sitting on this while they look at it, or does it have to go in ASAP?

ehelms:
We might could sit on this right up until release (or a user reports it) in hopes that a true fix is found.

ehelms:
Closing due to new qpid-dispatch packages that should address this issue.

PR #42, opened 2015-04-15
Remove use of functions from puppet-common in favor of stdlib.

ehelms:


ekohl:
Is this still relevant? It needs a rebase in any case.

ehelms:
I don't think it is - I'll close it as I think it would be better off trying to get what is in puppet-common into stdlib or extlib.

PR #43, opened 2015-04-16
refs #6781 - provide way of restarting capsule services

stbenjam:


manifests/params.pp
```diff
@@ -11,6 +11,12 @@
   $foreman_oauth_key          = cache_data('oauth_consumer_key', random_password(32))
   $foreman_oauth_secret       = cache_data('oauth_consumer_secret', random_password(32))
 
+  $services_directory         = '/etc/katello/services.d'
+  $foreman_proxy_services     = ['foreman-proxy']
+  $pulp_services              = ['qdrouterd', 'qpidd', 'pulp_celerybeat', 'pulp_workers',
+                                  'pulp_resource_manager', 'httpd']
```
jlsherrill: any thoughts on dhcp or dns ?   I could see arguments for and against having those also drop a services.d file

stbenjam: Yea, that's the problem, some proxy providers manage services not present on the capsule, others represent local services that we could restart.

ehelms:
@stbenjam is this PR still needed?

PR #44, opened 2015-05-05
fixes #10387 - set apache certs properly on crane module

jlsherrill:
due to an ordering issue the apache certs were not being set properly
splitting up this class would likely make the fix more declarative (as this is
a fairly imperative fix).  I have written a unit test to make sure we do not
regress.


ehelms:
APT

PR #47, opened 2015-05-26
Refs #10616 - use new pulp child ssl fragment for gpg key proxy

jlsherrill:


jlsherrill:
Requires https://github.com/Katello/puppet-pulp/pull/66

jlsherrill:
[test]

ehelms:
What do you think about a test that ensures this gets called with this content file if $pulp is true?

jlsherrill:
@ehelms added test, not sure how to test the content, as it is a rendered template as a param value, any thoughts?

spec/classes/capsule_spec.rb
```diff
@@ -45,6 +45,8 @@
 
     it { should contain_class('crane').with( {'key' => '/etc/pki/katello/private/katello-apache.key',
                                               'cert' => '/etc/pki/katello/certs/katello-apache.crt'} ) }
+
+    it { should contain_pulp__child__fragment('gpg_key_proxy') }
```
ehelms: I think you can do `with` here:

```
it 'should contain pulp gpg proxy' do
  should contain_pulp_child_fragment('gpg_key_proxy').with({
    :ssl_content => template('capsule/_pulp_child_gpg_proxy.erb'
  })
end
```

jlsherrill:
@ehelms updated

ehelms:
ACK

PR #48, opened 2015-06-02
Fixes #10675 - templates plugin enabled by default

lzap:
Otherwise indirect provisioning does not work. The template renders fine.


iNecas:
ACK

stbenjam:
It was off for a reason... this now turns the templates feature on the Katello, too, which (1) makes firewall rules incomplete (2) unnecessarily proxies unattended through the integrated capsule.

PR #50, opened 2015-07-07
refs #10675 - templates plugin should not be enabled by default

stbenjam:
Reverts Katello/puppet-capsule#48.  The problem is this now turns the Templates proxy on for Katello, too, so it unnecessarily proxies the unattended URL's, and makes our firewall docs incorrect about what ports go where.

It could be added to the capsule-installer command here: https://github.com/Katello/katello-installer/blob/master/hooks/post/10-post_install.rb#L63

Or if the capsule puppet module can have a way to know if its integrated or not, it can turn these things on by default.

Thoughts?

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -77,7 +77,7 @@ class capsule::params {
   $freeipa_remove_dns            = $foreman_proxy::params::freeipa_remove_dns
 
   # Templates proxy
-  $templates                     = true
+  $templates                     = false
 
   $register_in_foreman = false
   $certs_tar = undef


stbenjam:
Failing due to incompatibilities with newest pulp module :-\

ehelms:
This should probably be locked to 0.1-pulp right now

ehelms:
Does this mean that by default a Capsule won't be deployed in 'isolation' mode?

stbenjam:
Right, the original bug is good it should be default, but we need to make sure we're not affecting the Katello.

Users have to copy and paste the command for capsule-installer anyway to get the oauth tokens, so I was thinking of putting --templates=true there as a default.

Or, we need a reliable way in puppet-capsule to know if we're on a katello or not.

ehelms:
I think I see -- this is a case where the "default" capsule should have different behavior than a capsule out in the wild? Perhaps we need a way of denoting that so that we can make assertions or set values based upon whether its a special capsule that is deployed on the same server?

stbenjam:
Yea, any ideas what we can do?

@ekohl is `defined(Class['katello'])` safe to use here you think?  In the installer katello always comes before capsule (https://github.com/Katello/katello-installer/blob/master/config/katello-installer.yaml#L13), but I know there's some implications of puppet's `defined()` method

ehelms:
We could introduce a 'default' or 'on_server' type of parameter that we use. In the installer, we can set that for example in the answers file for the server installation.

stbenjam:
Ah ok, so something like this + https://github.com/stbenjam/katello-installer/commit/8f857fa190608308cc05b898863ea791dbbd350e

ehelms:
Sorry for the delay -- yea that example looks good to me

stbenjam:
Alright, updated, ready for review again.  I'll open the installer PR once this is merged.

@ekohl Thanks the pick is cleaner, and works. Add a spec to confirm.  I would generally agree about default, but it's only got one meaning in the context of a Katello capsule. I could call it `default_capsule` but that's redundant.

ehelms:
I think I might be swinging more into @ekohl camp here, while `default_capsule` seems redundant, its very obvious compared to simply `default`.

ekohl:
How about something more explicit? ```shared_server``` or ```same_server```? Though I wonder why a separate parameter is needed. Can't the installer set templates to false by default and only set it to true on the split server profile?

stbenjam:
> Can't the installer set templates to false by default and only set it to true on the split server profile?

@ekohl Ha! Yes.. yes it can.

stbenjam:
Not needed anymore, it can stay on, and I turn it off for Katello here: https://github.com/Katello/katello-installer/pull/243

Thanks @ekohl :-)

ekohl:
No problem :)

PR #54, opened 2015-09-02
Fixes #11660 - use default CA for client certificates verification

iNecas:
The katello_server_ca_cert is not meant to verify client certificates
(although it seems to work with all-in-one certificates).


ehelms:
ACK

PR #58, opened 2015-09-24
fixes #11958: fix node configuration

bbuckingham:
puppet-pulp was recently updated to introduce some 'node' parameters.
These parameters need to be set in the puppet-capsule, to ensure that
the /etc/pulp/nodes.conf is configured properly.


bbuckingham:
Note: without this, syncs from the Katello server to the Capsule will fail.

ehelms:
ACK

PR #61, opened 2015-10-21
Additional updates to support puppet-pulp 1.0.0

bbuckingham:
This commit contains some additional changes to support puppet-pulp 1.0.0
on a capsule.  This includes setting various pulp parameters to the
default values that katello/capsule had previously as well as inclusion
of the httpd conf files.


bbuckingham:
Note: this PR also requires https://github.com/Katello/puppet-pulp/pull/109, which addresses a possible error with pulp-manage-db

manifests/config.pp
```diff
@@ -0,0 +1,19 @@
+# Capsule Config
+class capsule::config {
+
+  file {'/etc/httpd/conf.d/pulp.conf':
```
ehelms: We should wrap these delcarations in an `if $pulp` check

bbuckingham:
@ehelms, updated

ehelms:
ACK

PR #62, opened 2015-10-26
fixes #12000 - add support for remote execution ssh plugin

stbenjam:
This is fugly, I don't even know if its worth including and perhaps waiting until foreman-proxy is at the top level.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -2,246 +2,259 @@
 #
 # === Parameters:
 #
-# $parent_fqdn::                    fqdn of the parent node. REQUIRED
+# $parent_fqdn::                        fqdn of the parent node. REQUIRED
 #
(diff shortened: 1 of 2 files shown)


ehelms:
Only ugly due to formatting change, I'd be OK with this if it means people playing with it in nightlies. I'll let @bbuckingham comment as well since he is handling the change you mention.

stbenjam:
Yea the formatting is nasty, but I meant wrapping the plugin in the capsule module itself isn't very nice either

bbuckingham:
I am Ok with the ugliness to make the functionality available.  I am sure it will mean some merge conflicts when the scenario-based changes go in, but that is to be expected.

ehelms:
ACK

PR #64, opened 2015-11-09
refs #10533 - initial changes to support decoupling puppet-foreman_proxy from puppet-capsule

bbuckingham:
MODIFIED description:

decouple puppet-foreman_proxy from puppet-capsule

The changes in this PR are an initial step towards enabling katello and the capsule to be deployed as a kafo 'scenario' using the foreman-installer.  It will also contain the configs to support installing using katello-installer and capsule-installer, for the near-term.

The following is an example of using the scenario-based installer to install a capsule using this module:
```
foreman-installer --scenario capsule \
                  --capsule-parent-fqdn                 "katello.example.com"\
                  --foreman-proxy-register-in-foreman   "true"\
                  --foreman-proxy-registered-proxy-url  "https://capsule.example.com:9090"\
                  --foreman-proxy-oauth-consumer-key    "QSJ2qCGV3zWyohCJfKxQBDCCjceBiTMy"\
                  --foreman-proxy-oauth-consumer-secret "DGKBJjXWfkyroiyDHNhYAFn9uTnUcJxa"\
                  --foreman-proxy-foreman-base-url      "https://katello.example.com"\
                  --foreman-proxy-trusted-hosts         "katello.example.com"\
                  --foreman-proxy-trusted-hosts         "capsule.example.com"\
                  --capsule-pulp-oauth-secret           "bQ5zUPVd2qVDY987D4VtD6Va9pDj4Y7J"\
                  --capsule-certs-tar                   "~/capsule.example.com-certs.tar"\
                  --capsule-puppet                      "true"\
                  --capsule-puppetca                    "true"\
                  --capsule-pulp                        "true"
```

The following is an example of using the existing capsule-installer to install a capsule using this module:
```
capsule-installer --register-in-foreman         "true"\
                  --registered-proxy-url        "https://capsule.example.com:9090"\
                  --oauth-consumer-key          "pvUutwvPU3Rni5RWqisL6Tx4fAF5yDSa"\
                  --oauth-consumer-secret       "JTGnd3fx6GB4uki5LtNPFU6cCSq9NrFk"\
                  --foreman-base-url            "https://katello.example.com"\
                  --trusted-hosts               "katello.example.com"\
                  --trusted-hosts               "capsule.example.com"\
                  --parent-fqdn                 "katello.example.com"\
                  --pulp-oauth-secret           "3HNmounKP6mbQYthkvkw2GXV5toxeFsu"\
                  --certs-tar                   "~/capsule.example.com-certs.tar"\
                  --puppet                      "true"\
                  --puppetca                    "true"\
                  --enable-foreman-proxy-plugin-pulp\
                  --enabled                     "false"\
                  --pulpnode-enabled            "true"
```

The following is an example of using the existing katello-installer to install using this module:
```
katello-installer
```

Note: In order to enable pulp and leverage parameters provided by the puppet-foreman_proxy,
the following arguments are now used : --enable-foreman-proxy-plugin-pulp (to enable the pulp plugin),
--enabled (do not configure as a pulp master), --pulpnode-enabled (configure as a pulp node)


bbuckingham:
Note: this PR is dependent on the changes to kafo included with https://github.com/theforeman/kafo/pull/117.

stbenjam:
Can't these be moved to foreman-proxy as well?

```
--capsule-parent-fqdn "katello.example.com"\
--capsule-puppet "true"\
--capsule-puppetca "true"\
--capsule-pulp "true"
```

.fixtures.yml
```diff
@@ -13,6 +13,7 @@ fixtures:
     foreman:       "git://github.com/theforeman/puppet-foreman"
```
ehelms: We should be able to get rid of a lot of the fixtures present here since they would no longer be referenced with the removal of foreman_proxy

manifests/init.pp
```diff
@@ -328,75 +148,6 @@
...
-    class { '::foreman_proxy::plugin::remote_execution::ssh':
-        enabled           => $remote_execution_ssh,
-        generate_keys     => $remote_execution_ssh_generate_keys,
-        ssh_identity_dir  => $remote_execution_ssh_identity_dir,
-        ssh_identity_file => $remote_execution_ssh_identity_file,
-        require           => Class['::foreman_proxy'],
-    }
-  }
-
   if $pulp {
```
ehelms: Do we still need this parameter if the assumption is this installs and configures Pulp on a smart proxy?

.fixtures.yml
ekohl: But it's still in a precondition, so you can't remove them yet.

ehelms: Just Foreman or all of them? I was referencing auditing the list as a whole (e.g. tftp, xinetd, dhcp, dns, foreman_proxy etc.)

ekohl: I think you may not require all repositories, but I suspect you will still need most of them. Doing a full check can't hurt though :)

manifests/init.pp
bbuckingham: @ehelms, Since the capsule-installer today provides the option to enabled/disable pulp, I would assume that we should maintain that for a capsule scenario.

ehelms: what is a capsule scenario without Pulp in the new world?

bbuckingham: Perhaps for users that are used to the old world that use the 'capsule' scenario to manage all of their proxy services (pulp, puppet, dhcp, dns...etc).  I can certainly remove it, but seems like it could cause confusion.

.fixtures.yml
bbuckingham: I was able to remove dhcp/dns; however, removing any additional ones at this time does result in test failures.

ehelms:
This got the errors fixed up for me:

```
diff --git a/manifests/init.pp b/manifests/init.pp
index f2ac476..2be2ab2 100644
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -79,6 +79,7 @@ class capsule (
   validate_present($capsule::parent_fqdn)
 
   $pulp = $::foreman_proxy::plugin::pulp::pulpnode_enabled
+
   if $pulp {
     validate_present($pulp_oauth_secret)
   }
diff --git a/spec/classes/capsule_spec.rb b/spec/classes/capsule_spec.rb
index 89f2218..1b6fe33 100644
--- a/spec/classes/capsule_spec.rb
+++ b/spec/classes/capsule_spec.rb
@@ -13,10 +13,6 @@ describe 'capsule' do
       }
     end
 
-    let(:pre_condition) do
-      ['include foreman_proxy']
-    end
-
     it { should contain_class('capsule::install') }
   end
 
@@ -33,14 +29,13 @@ describe 'capsule' do
 
     let(:params) do
       {
-        :pulp              => true,
         :pulp_oauth_secret => 'mysecret',
         :qpid_router       => false
       }
     end
 
     let(:pre_condition) do
-      ['include foreman_proxy']
+      "class {'foreman_proxy::plugin::pulp': pulpnode_enabled => true}"
     end
 
     it { should contain_class('crane').with( {'key' => '/etc/pki/katello/private/katello-apache.key',
```

bbuckingham:
@ehelms, Thanks for the test updates!  Applied them with a rebase to the PR.

ehelms:
@bbuckingham do you still consider this in a DNM state? I am feeling fairly confident in it to consider merging and revving the version. I believe this new version would still get imported into katello-installer, so we could consider putting in a version lock in Puppetfile or only merging this once https://github.com/Katello/katello-installer/pull/283 is ready -- which I think it is if we are doing the migration PR after the fact?

bbuckingham:
@ehelms, I do not have any changes pending on this particular PR.  If we find something after it is in place, we can create separate PRs.  We will need the changes that you referenced for katello installer (https://github.com/Katello/katello-installer/pull/283) as well as the kafo migration changes that martin is working on to be fully functional (for existing installs).

ehelms:
We can either bump the version here, or after this I can use blacksmith to do so.

bbuckingham:
@ehelms, We can bump it however you prefer.

ehelms:
ACK - I'll do a bump after we merge this

bbuckingham:
@ehelms, Thanks!  Will merge now. :)

PR #68, opened 2016-01-13
Fixes #13172 - New pulp params for pulp.yml and pulpnode.yml

xprazak2:
New parameters that will be used in proxy observability.

### DIFF
--- a/templates/pulp.yml
+++ b/templates/pulp.yml
@@ -1,4 +1,7 @@
 ---
 :enabled: <%= @pulp_master %>
 :pulp_url: https://localhost/pulp
-
+# Path to pulp, pulp content and mongodb directories
+# :pulp_dir: /var/lib/pulp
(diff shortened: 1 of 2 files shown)


ehelms:
I think you will want to do this in the main puppet-foreman_proxy that has the Pulp plugin. If you look in this folder (https://github.com/theforeman/puppet-foreman_proxy/tree/master/templates/plugin) you will see them there. I dont' recall the historical reason for it, but we are moving away from these local ones to use the smart proxy plugin to dictate. You can see they are removed in our work to separate puppet-foreman_proxy from puppet-capsule -- https://github.com/Katello/puppet-capsule/pull/64/files#diff-60ae41fd0a31977447947f59940ee9a4L306

xprazak2:
OK, I will submit new params to puppet-foreman_proxy and close this. Thank you!

PR #70, opened 2016-02-03
include apache class as it is used to fetch apache version

jlsherrill:


manifests/init.pp
```diff
@@ -72,6 +72,7 @@
   $qpid_router_broker_port   = $capsule::params::qpid_router_broker_port,
 ) inherits capsule::params {
 
+  include ::apache
```
ekohl: This and ```$apache::version``` can be moved to ```apache::vhost``` (line 138) so the're only included if needed.

jlsherrill: if i do that i need to also include those two lines in config.pp (as some config files there also reference apache_version).  Which is preferred?

ekohl: I'm leaning to getting rid of ```capsule::config``` and move it to init.pp. Might as well get rid of ```capsule::install``` while you're at it. They both appear to be leftovers from when this module did a lot more than it used to.

@ehelms do you agree?

ehelms: For this change I suppose I can live with that -- I really need to bite the bullet and take the time to split this up more cleanly for the various concerns it tackles.

ekohl: I agree that this has become much more of a profile module and we complete that transition.

jlsherrill:
Updated, i wasn't sure if the notify was needed that was there before.  It didn't seem like it was.

ekohl:
It looks like you forgot to remove the ```::config``` class, other than that it's looking good.

jlsherrill:
Doh!  Thanks @ekohl

ehelms:
ACK

PR #71, opened 2016-02-09
Refs #13607 - Removed pulp.conf

parthaa:
The template is now removed and centralized in puppet-pulp

check https://github.com/Katello/puppet-pulp/pull/124


manifests/init.pp
```diff
@@ -137,14 +137,6 @@
...
 
-    file {'/etc/httpd/conf.d/pulp.conf':
-      ensure  => file,
-      content => template('capsule/pulp.conf.erb'),
-      owner   => 'root',
-      group   => 'root',
-      mode    => '0644',
-    }
-
     file {'/etc/httpd/conf.d/pulp_nodes.conf':
```
ekohl: What about this? Should this be moved as well?

ehelms: Should line 138 also be removed now?

ekohl: No, still used in httpd_pub.erb.

ehelms:
ACK - thanks @parthaa !

PR #81, opened 2016-04-07
Fixes #14534 - ensure we install client-bootstrap

mccun934:
The client-bootstrap lands in /var/www/html/pub and is needed on the
capsule for clients without access top to the Foreman server


ehelms:
ACK - thanks @mccun934 !

PR #86, opened 2016-06-07
Updating to support 3.X proxy modules

jlsherrill:


jlsherrill:
Didn't bump the minimum as technically there was no module breakage, we just need the newer version

stbenjam:
APT

PR #87, opened 2016-06-21
refs #15217 - puppet 4 support

stbenjam:


ekohl:
It fails because the ```$::rubyversion``` fact is not available in the tests. In theforeman we use ```rspec-puppet-facts``` to set all the facts and in general I'm pretty happy with that. I'm sure other of your puppet 4 PRs fail on the same problem.

stbenjam:
Green

ekohl:
ACK

PR #90, opened 2016-08-04
Pass through server implementation parameter to puppet class

stbenjam:
We need to expose this for the upgrade from puppet 3 to 4 (instructions in progress @ bottom of http://pad-katello.rhcloud.com/p/puppet4)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -24,6 +24,9 @@
 #                                       the apache vhost to set up a proxy for all
 #                                       certificates pointing to the value.
 #
+# $puppet_server_implementation::       Puppet master implementation, either "master" (traditional
+#                                       Ruby) or "puppetserver" (JVM-based)
+#
 # $reverse_proxy::                      Add reverse proxy to the parent
 #                                       type:boolean
 #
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -49,31 +52,33 @@
...
+  $certs_tar                    = $capsule::params::certs_tar,
+  $pulp_master                  = $capsule::params::pulp_master,
+  $pulp_admin_password          = $capsule::params::pulp_admin_password,
+  $pulp_oauth_effective_user    = $capsule::params::pulp_oauth_effective_user,
+  $pulp_oauth_key               = $capsule::params::pulp_oauth_key,
+  $pulp_oauth_secret            = $capsule::params::pulp_oauth_secret,
+
+  $puppet                       = $capsule::params::puppet,
+  $puppet_ca_proxy              = $capsule::params::puppet_ca_proxy,
+  $puppet_server_implementation = $capsule::params::puppet_server_implementation,
```
ekohl: https://github.com/Katello/puppet-capsule/pull/90/files?w=1 for the lazy :)

manifests/params.pp
```diff
@@ -19,6 +19,19 @@
...
 
+  if versioncmp($::puppetversion, '4.0') < 0 {
+    $aio_package      = false
+  } elsif $::rubysitedir =~ /\/opt\/puppetlabs\/puppet/ {
+    $aio_package      = true
+  } else {
+    $aio_package      = false
+  }
+
+  $server_implementation = $aio_package ? {
```
ekohl: If we leave this as undef, will that work as well?

stbenjam: If you pass an undef variable to a class, does it take the default instead?

ekohl: It does. I think the installer handles this as well

beav:
I was able to upgrade from puppet 3 to 4 after applying this patch and running through the upgrade procedure.

ekohl:
ACK

beav:
@ekohl thanks!

@stbenjam is away, merging on his behalf.

PR #91, opened 2016-08-23
refs #10283 - mark parameters advanced

stbenjam:
Anything other than "Parameters" are not shown by default in Kafo help.
This makes the installer help shorter, and reduces the chance users will
change options that really shouldn't be changed unless you really really
really know what you're doing.


ehelms:
ACK

PR #92, opened 2016-08-29
Only disable passenger on pub if passenger is present

stbenjam:
As puppet 4 no longer uses passenger, passenger isn't present
and this errors out:

AH00526: Syntax ierror on line 31 of /etc/httpd/conf.d/05-capsule.conf:
Invalid command 'PassengerEnabled', perhaps misspelled or defined by a
module not included in the server configuration

### DIFF
--- a/templates/httpd_pub.erb
+++ b/templates/httpd_pub.erb
@@ -1,7 +1,9 @@
 alias /pub /var/www/html/pub
 
 <Location /pub>
-  PassengerEnabled off
+  <IfModule mod_passenger.c>
+    PassengerEnabled off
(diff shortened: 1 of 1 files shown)


beav:
This is the same as a similar change in the Katello module, APJ. I do not see `PassengerEnabled` mentioned anywhere else, so this change should cover all uses of it.

ekohl:
https://github.com/Katello/puppet-certs/pull/102 should help with the build failure.

ehelms:
@stbenjam appears to be green now

ekohl:
ACK

PR #99, opened 2016-10-17
Fixes #16972 - add ssl.conf for mod_ssl pkg bug

chris1984:
Talked to Eric and Justin and since we cant not atm get the fix into upstream apache, this is the way to go for our downstream issue.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -145,6 +145,14 @@ class capsule (
         require => Class['pulp'],
       }
     }
+    # adding ssl.conf back to resolve Katello issue #16972 - http://projects.theforeman.org/issues/16972
+    file {'/etc/httpd/conf.d/ssl.conf':
+      ensure  => file,
+      content => '# Managed by Puppet',
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0644',
+    }
 
     class { '::crane':
       cert    => $certs::apache::apache_cert,
--- a/spec/classes/capsule_spec.rb
+++ b/spec/classes/capsule_spec.rb
@@ -16,10 +16,12 @@ describe 'capsule' do
           }
         end
 
-        let(:pre_condition) do
-          "class {'foreman_proxy::plugin::pulp': pulpnode_enabled => true}
+
+      let(:pre_condition) do
+         "class {'foreman_proxy::plugin::pulp': pulpnode_enabled => true}
          class {'apache': apache_version => '2.4'}"
         end
+
 
         it { should contain_class('crane').with( {'key' => '/etc/pki/katello/private/katello-apache.key',
                                                   'cert' => '/etc/pki/katello/certs/katello-apache.crt'} ) }


lpramuk:
On puppet4 environments live elsewhere in ```/etc/puppetlabs/code/environments``` instead of ```/etc/puppet/environments``` that's why Travis on Puppet4 is failing

chris1984:
@lpramuk thanks for the update, I talked to @stbenjam and he had me submit a PR to fix it in another repo. Once he finishes testing then i will resubmit to kick off the tests and it should work.

manifests/init.pp
```diff
@@ -146,6 +146,14 @@
       }
     }
 
+    file {'/etc/httpd/conf.d/ssl.conf':
+      ensure  => file,
+      content => '# Managed by Puppet',
```
ehelms: I think it could be worth putting in here either the reason for this or a link to either the puppet-apache issue or our issue to help anyone in the future who asks why this is the way it is to know.

chris1984: @ehelms ack will add this and resubmit once stbenjam finishes with his testing and pulls in the pr I put in yesterday adding the env variable so the tests will go through with this pr.

spec/classes/capsule_spec.rb
```diff
@@ -16,11 +20,13 @@
           }
         end
 
-        let(:pre_condition) do
-          "class {'foreman_proxy::plugin::pulp': pulpnode_enabled => true}
+
+      let(:pre_condition) do
+         "class {'foreman_proxy::plugin::pulp': pulpnode_enabled => true, puppet_content_dir => '/etc/puppet/environments'}
```
stbenjam: Please remove this puppet_content_dir from the test

spec/classes/capsule_spec.rb
```diff
@@ -5,6 +5,10 @@
     context "on #{os}" do
       let(:facts) { facts }
 
+      let(:pre_condition) do
+         "class {'foreman_proxy::plugin::pulp': puppet_content_dir => '/etc/puppet/environments'}"
```
stbenjam: Ditto

Gemfile
```diff
@@ -33,5 +33,4 @@ gem 'json', '~> 1.0', {"platforms"=>["ruby_18", "ruby_19"], "groups"=>["test"]}
 gem 'json_pure', '~> 1.0', {"platforms"=>["ruby_18", "ruby_19"], "groups"=>["test"]}
 gem 'metadata-json-lint'
 gem 'rspec-puppet-facts', '>= 1.7'
-
```
ekohl: Please revert this change. It's managed by modulesync and just creates noise.

stbenjam:
I'm guessing the puppet 4 vs. puppet 3 is an ordering issue of resources, but `::apache` is getting included in puppet-capsule (not sure why, probably for params but not really important).  That ends up calling `::apache:mod::ssl`.  Which already has an ssl.conf resource.  So, you're getting a duplicate file error when we come along and it's already defined.  But only on puppet 4 because of how the resources get ordered.

Ideally, mod_ssl packaging would allow you to define Listen 443 in a different file and not replace ssl.conf if it was missing.  We should try to find a solution from them for this.

stbenjam:
Ah, not an ordering issue. It fails on RHEL6, because puppetlabs-apache uses /etc/httpd/conf.d/ssl.conf, but on el6, it uses /etc/httpd/conf.modules.d/

ehelms:
Ultimately, mod_ssl package fix is the proper, best fix, but I am missing why we can't do this here for now to solve our users upgrade issues? If its an EL6 vs EL7 problem, we should be able to just wrap it in an if statement?

stbenjam:
The existing hook **already** solves the upgrade issue, which is where 99% of users hit the problem. Red Hat QE failed the bug because it doesn't solve every use case (a future update of httpd will replace the ssl.conf again, when the user has not run the --upgrade).

I don't really like carrying bug fixes in Katello for other software and we really far too often want to accept them.  "Temporary" workarounds end up staying here forever - :-1:  :-1:  :-1:  Especially in puppet modules, it's buried here and doesn't get a very close look often.

Anyway, we found another solution, it'll go in a post hook too which should make Red Hat QE happy and me less unhappy.  I feel like hooks are better for "temporary" things, and get looked at more often.

stbenjam:
And a good case in point of things that caused us pain later: the mongo PID file "fix" we hacked into Katello.  Nothing but headaches, it's also still in the katello-installer code even though I think EPEL fixed the problem in their packaging now.

ekohl:
Another thing to consider: why should it go in this module? puppet-foreman and puppet-puppet can also include apache and can hit the same problem. You can't define the resource multiple times so *if* (and that's a big if) you want to include a workaround it belongs in puppetlabs-apache.

stbenjam:
> You can't define the resource multiple times so if (and that's a big if) you want to include a workaround it belongs in puppetlabs-apache.

This is the source of the duplicate file on el6 (it uses conf.d for everything).
  https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/params.pp#L69-L71

But, I assume this is for a good reason, it looks much nicer on newer to have all the module configuration (especially since there's a lot of files) separate from the regular conf.d if it's possible.

I guess you could make an exception in ::apache::mod::ssl and have it put `ssl.conf` in conf_dir, but not sure if they'd accept that.

ekohl:
I'm not sure I fully understand the problem.

So the problem exists if ```apache``` is included but not ```apache::mod::ssl```? The latter manages ```ssl.conf``` and guarantees it's set up correct, but if that's not included but ```mod_ssl``` is installed (after puppet purged the config dir) then you end up with a duplicate listen definition. That would mean a possible workaround is to just add ```include ::apache::mod::ssl``` after ```include ::apache```.

On the other hand, if mod_ssl doesn't treat ```ssl.conf``` like a config file and always overwrites it on upgrade then a ```yum update``` will break it again.

ehelms:
I agree that temporary fixes are not great and we do tend to forget about them though we could do better at creating removal issues and assigning them to the release we want to update things in. However, we lack control over third party projects and when official updates go out so I'd rather help our users then shrug our shoulders because the proper fix isn't going in.

I was more a fan of this fix in a puppet module due to Puppet having better insurance of idempotency, to help prevent any weird thing users might do like.  Granted, this could be a false assurance I am telling myself exists.

@ekohl In all honesty, some of us have a history of things not getting in or getting in when we need them  on the Foreman side that makes us tend to put things where we have control over them. Not saying this is still true, but some things take a while to unlearn.

@stbenjam or @chris1984  have we filed an official issue and/or do we know the status with the mod_ssl package itself?

ekohl:
> @ekohl In all honesty, some of us have a history of things not getting in or getting in when we need them on the Foreman side that makes us tend to put things where we have control over them. Not saying this is still true, but some things take a while to unlearn.

I get that, but I'm trying to figure out when it is triggered. If I understand it correctly then the reason it's here and not in puppet-foreman or puppet-puppet that's because they will pretty much always include ```apache::mod::ssl```. With puppet-capsule you don't have that guarantee.

You can safely unconditionally include ```apache::mod::ssl``` in terms of puppet usage. The only downside is that you may have a slightly higher memory usage but I think that's negligible.

stbenjam:
@ekohl  Nope, that's not correct understanding of the problem.  ::apache:mod::ssl is included here via ::include apache, and in puppet-foreman indirectly by enabling an ssl vhost.

On rhel 6, it creates a file called /etc/httpd/conf.d/ssl.conf which includes module configuration. On el 7, that file is called /etc/httpd/conf.modules.d/ssl.conf. Which is why tests failed on this PR for el6.

The original source of this issue is the RPM package ships /etc/httpd/conf.d/ssl.conf that configures a vhost that listens on 443 -- **totally different than the puppet's ssl.conf purpose**. 

The problem is ssl.conf is added on every update of httpd package, and the ssl.conf in the rpm package duplicates the Listen 443 segment in the ports.conf puppetlabs-apache creates.  The RPM needs fixing to not put the file back if it was deleted (which happens on el7).

We added a hook a while ago that fixes the case where users upgraded katello and now can't start because they have 2 listen statements:  https://github.com/Katello/katello-installer/pull/383

That fixes *the vast vast majority* of cases but Red Hat QE is unhappy as I explained above that it doesn't fix every case (which SHOULD go to mod_ssl RPM).

The proposed additional host pook would fix it, or wrapping this in an operating system check would probably too. I don't really care anymore, do whatever.

ekohl:
@stbenjam right, so the original problem does happen on EL7 (which has 2 ssl.conf files, both in conf.d and conf.modules.d) but not on EL6 (which only has conf.d/ssl.conf).

Does the problem only show up on capsules or also on installs without a capsule?

stbenjam:
> @stbenjam right, so the original problem does happen on EL7 (which has 2 ssl.conf files, both in conf.d and conf.modules.d) but not on EL6 (which only has conf.d/ssl.conf).

Er, sorry, still not quite right. 

EL7 only has a ssl.conf in conf.modules.d.  The original ssl.conf in /etc/httpd/conf.d/ssl.conf is removed by puppetlabs-apache clearing out the directory.  It's very confusing to explain.

As I said, the ssl.conf's despite the same names are not doing the same thing - puppetlabs-apache is deploying an ssl.conf used to configure mod_ssl general settings.  The ssl.conf laid down by red hat includes a listen 443 to turn on ssl automatically, but it conflicts w/ ports.conf. It just so happens that the el6 ssl.conf is laid down in /etc/httpd/conf.d, so the rpm doesn't replace it since it was modified. It does on el7 since there is no file in /etc/httpd/conf.d/ssl.conf.

It happens on foreman, too. It's not just a katello problem.

It's an RPM bug.

ekohl:
Oh yes, that was my interpretation as well. I was talking about the EL7 RPM, not what puppetlabs-apache does with it.

IMHO pl-apache should deploy an empty ```conf.d/ssl.conf``` on EL7. At least until there's a mod_ssl RPM fix but even then you will probably have to carry that for a long time due to older versions. The bug is in how puppetlabs-apache and mod_ssl interact. Solving it in the installer or any other puppet module is the wrong place because you have to repeat the workaround multiple times.

stbenjam:
If we'd fix it in puppetlabs-apache, just changing this location on el7 is enough:

    https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/mod/ssl.pp

Since that ssl.conf can replace the rpm ssl.conf just like el6 does.  It's still hacky imho, as the puppet ssl.conf is module configuration that belongs now in conf.modules.d.

lpramuk:
@stbenjam I dont think httpd ppl will fix it. Currently we have:
```%config(noreplace) %{_sysconfdir}/httpd/conf.d/ssl.conf``` that is if file exist update will create it as *.rpmnew, if it doesnt exist it gets created with original name.

If you propose to have ```%config(missingok)``` the httpd team will refuse it eventually as it's not really ok if the file is missing from their point of view, without it though mod_ssl is installed HTTPS won't work, ```missingok``` is not ok.

Also many teams that use httpd for their products are able to handle current situation with  ```noreplace``` just fine. I dont think pushing at httpd is not good solution here. It's not  mod_ssl RPM bug.

stbenjam:
> If you propose to have %config(missingok) the httpd team will refuse it eventually as it's not really ok if the file is missing from their point of view, without it though mod_ssl is installed HTTPS won't work, missingok is not ok.

This is just not true - you can put the settings in any file.  It doesn't need to be called 'ssl.conf', and indeed on Foreman it's not.

> Also many teams that use httpd for their products are able to handle current situation with noreplace just fine. I dont think pushing at httpd is good solution here. It's not mod_ssl RPM bug. 

puppetlabs-apache configures Apache across multiple distributions, and uses some common puppet patterns like fragments to build up a file called port that has all the places where Listen statements go.  We are free to do that by the way Apache works.  The fact mod_ssl is imposing the requirement that ssl Listen statement goes in ssl.conf is a bug, and it's 100% solvable in the package spec.  I'm not 100% sure if missingok does what I want, I think it doesn't, but you can check in %pretrans if the file was removed and only place it on the system in %posttrans.  That might not be the right solution, but instead of imaging what the team will or won't say, open a bug and explain the problem.

The bottom line is: Katello or Foreman is NOT the right place to fix this.

You could propose a fix to puppetlabs-apache (although I'm not sure I really agree that's the right place), or we try to fix it in mod_ssl.

stbenjam:
FWIW, here's a proposed change to puppetlabs-apache: https://github.com/puppetlabs/puppetlabs-apache/pull/1543

I still think it's less than ideal than coming up with a scheme in mod_ssl packaging to not remove the ssl.conf if it was explicitly removed by a user.

lpramuk:
Thanks for explanation. 

I agree on fixing puppetlabs-apache, if it doesn't want to use ssl.conf at all it is upto this module to configure it properly - i.e. leave it empty. 
(by leaving it empty the module say that ssl was configured elsewhere, i.e not by ssl.conf)

Can ```rpm``` handle it (explicitly removed conf)? I know that ```apt``` handles it.

PR #104, opened 2016-11-02
Use crane from puppet-pulp

stbenjam:
Needs https://github.com/Katello/puppet-pulp/pull/170


ehelms:
Looks good pending puppet-pulp change

PR #105, opened 2016-11-10
Crane needs configuration of SSL certs separate from pulp

stbenjam:
On the Katello, pulp is `manage_apache => false` and uses it's own
CA.  On a capsule, pulp is `manage_apache => true` and uses our
apache certs.  So, we need to configure crane directly to use
apache certs in all cases.  Currently on a Katello it's using the Pulp
certs which causes all kinds of problems.

This fixes BATS.  

Eventually we need to figure out how to standardize
the pulp configurations so they're more similar.  If we do that, we can
go back to configuring crane through the pulp top class.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -145,6 +145,13 @@ class capsule (
         require => Class['pulp'],
       }
     }
+
+    class { '::pulp::crane':
+      cert    => $certs::apache::apache_cert,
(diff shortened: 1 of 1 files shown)


ehelms:
We want to move to puppet-foreman_proxy_content configuring services anyway.

jlsherrill:
ACK from me as well

PR #106, opened 2016-11-21
refs #17366 - rename to puppet-foreman_proxy_content

stbenjam:


ehelms:
Should we reset the versioning since its a rename?

stbenjam:
Sure,  that works.  I think we could really do anything, I was originally thinking to call it 5.0.  But do you have any preference? 1.0? 0.1? 0.0.1?

ehelms:
I'm good with 1.0 unless you feel it should go through a bit of non-semantic versioning before saying its "1.0" worthy.

PR #108, opened 2016-12-16
refs #17714 - Remove puppet from foreman_proxy_content

stbenjam:


manifests/init.pp
```diff
@@ -217,33 +203,21 @@
...
-      server_ca_proxy             => $puppet_ca_proxy,
-      server_implementation       => $puppet_server_implementation,
-      additional_settings         => {
-                                        'disable_warnings' => 'deprecations',
-      },
-    }
+  # We can't pull the certs out to the top level, because of how it gets the default
+  # parameter values from the main ::certs class.  Kafo can't handle that case, so
+  # it remains here for now.
+  class { '::certs::puppet':
```
ehelms: To be more specific, this block here -- https://github.com/Katello/puppet-foreman_proxy_content/pull/108/files#diff-60ae41fd0a31977447947f59940ee9a4R213

stbenjam: Yea I think so

manifests/init.pp
```diff
@@ -217,33 +203,21 @@
     }
   }
 
-  if $puppet {
+  # We can't pull the certs out to the top level, because of how it gets the default
+  # parameter values from the main ::certs class.  Kafo can't handle that case, so
+  # it remains here for now.
+  if defined(Class['puppet']) {
```
ekohl: This is dependent on the compile order so I'm not sure this is the best solution to detect if it was enabled.

manifests/init.pp
```diff
@@ -217,33 +203,21 @@
...
-      server_config_version       => '',
-      server_enc_api              => 'v2',
-      server_ca_proxy             => $puppet_ca_proxy,
-      server_implementation       => $puppet_server_implementation,
-      additional_settings         => {
-                                        'disable_warnings' => 'deprecations',
-      },
     }
+
+    Class['certs::puppet'] -> Class['puppet']
```
ekohl: This could also go in the above class as a ```before```.

manifests/init.pp
```diff
@@ -217,33 +203,21 @@
...
   }
 
   if $certs_tar {
     certs::tar_extract { $foreman_proxy_content::certs_tar: } -> Class['certs']
     Certs::Tar_extract[$certs_tar] -> Class['certs::foreman_proxy']
+    Certs::Tar_extract[$certs_tar] -> Class['certs::puppet']
```
ekohl: Won't this give a problem with strict definitions when puppet is not enabled? Maybe in the above puppet block you can chain ```Class['certs'] -> Class['certs::puppet']``` and since you already chain the extract before you implicitly get the same chaining.

stbenjam: It's not even possible to install Foreman without Puppet, nevermind Katello, so not sure we should really be addressing that here.

stbenjam:
@ekohl Ready for another look, I put the `$puppet` param back so if a user wants to try to disable puppet they can.

manifests/init.pp
```diff
@@ -218,27 +209,13 @@
...
-      server_environments_owner   => 'apache',
-      server_config_version       => '',
-      server_enc_api              => 'v2',
-      server_ca_proxy             => $puppet_ca_proxy,
-      server_implementation       => $puppet_server_implementation,
-      additional_settings         => {
-                                        'disable_warnings' => 'deprecations',
-      },
-    }
+      # We can't pull the certs out to the top level, because of how it gets the default
```
ekohl: Indenting is a bit off here

manifests/init.pp
```diff
@@ -218,27 +209,13 @@
...
+      # parameter values from the main ::certs class.  Kafo can't handle that case, so
+      # it remains here for now.
+      class { '::certs::puppet':
+        hostname => $foreman_proxy_fqdn,
+        before   => Class['puppet'],
```
ekohl: Should it be notify so the puppet server will restart on new certificates?

stbenjam:
@ekohl Updated

PR #113, opened 2017-02-13
[WIP] fixes #17367 - backwards-compatible rhsm proxy on main vhost

stbenjam:
This issue has been open for a while, and I've been thinking about how to do
it in a way that doesn't break users.  I think I have one, although this
is still a WIP.  puppet-pulp will need changes, and we'll need
to get a new puppetlabs-apache with https://github.com/puppetlabs/puppetlabs-apache/commit/6ff47c4e5a5fe84439e116820db2d34a96e125e7.

This moves the RHSM proxy to the main Apache vhost, so the URL's
for a Foreman proxy and a Katello both use the same port (443 by
default).  But we can't just break existing capsules where many
clients could be pointed to 8443 still.

In order to be backwards compatible with previous installations which
hosted a reverse proxy on 8443, it allows configuration of multiple
ports for the main Apache vhost to listen on.

New installs would default to 443 for everything.

Users have a couple of options for upgrade scenario:

(1)  The default upgrade would require no changes on their part:

        - Separate reverse proxy vhost goes away
        - Pulp vhost listens on 8443, too (installer migration sets this)
        - Clients need no updates

(2) They can also opt to continue listening on 8443, but update the RPM
to point to 443 so new clients point to 443. They can move existing
clients at their convenience and eventually discontinue 8443.

(3) Or they can just stop listening on 8443 entirely and use 443, and
update all of their clients with a new RPM.


stbenjam:
@ehelms Would appreciate your thoughts on the approach - anything I'm missing? 

The PR as is doesn't quite work yet without some fiddling, I need to make some changes to puppet-pulp too.

ehelms:
Are the changes to puppet-pulp so you can pass the pulp_ports variable? You might add that here even if it doesn't work just to see the full workflow.

stbenjam:
Updated

manifests/init.pp
```diff
@@ -36,15 +36,18 @@
...
 #                                       type:Optional[String]
 #
-# $reverse_proxy::                      Add reverse proxy to the parent
-#                                       type:Boolean
+# $pulp_ports::                         Ports for the Pulp apache vhost to listen on
+#                                       type:Array[Integer[0, 65535]]
 #
-# $reverse_proxy_port::                 Reverse proxy listening port
-#                                       type:Integer[0, 65535]
+# $rhsm_proxy::                         Add a proxy rule for Red Hat Subscription Manager to the Pulp vhost
```
ehelms: I'm not sure this makes sense as an optional parameter with how we deploy configuration currently. Unless you use pure HTTP based Pulp repositories that you configure access to yourself, we have to couple deploying the RHSM proxy with having Pulp deployed to a proxy. Another way to look at it is, the katello-consumer RPM deployed to the proxy is deployed with this assumption.

ekohl:
There's at least a merge conflict.

ekohl:
@stbenjam any plan to revisit this?

stbenjam:
I'd like to, but probably not until I'm back full time, so we can close this for now.

PR #114, opened 2017-03-18
reload foreman-proxy service when cert changes

timogoebel:
This commit reload foreman-proxy service when the cert changes. This has
previously been done in puppet-certs. To reduce dependencies of
puppet-certs, moving it here.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -130,7 +130,7 @@ class foreman_proxy_content (
   class { '::certs::foreman_proxy':
     hostname => $foreman_proxy_fqdn,
     require  => Package['foreman-proxy'],
-    before   => Service['foreman-proxy'],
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -130,6 +130,7 @@
   class { '::certs::foreman_proxy':
     hostname => $foreman_proxy_fqdn,
     require  => Package['foreman-proxy'],
+    notify   => Service['foreman-proxy'],
     before   => Service['foreman-proxy'],
```
ekohl: notify implies a before relation so that can be removed.

timogoebel:
@ekohl : Thanks. Updated. Also fixed a puppet-lint issue.

ehelms:
Thanks @timogoebel !

PR #115, opened 2017-03-24
Fixes #19016 - change qpid to localhost

Klaas-:


Klaas-:
Tests are broken for Puppet 3.5 (not by my change as far as I can tell)

Klaas-:
I think a dependency or 2nd level dependency stopped supporting puppet3 :)

ekohl:
@Klaas- that's very possible, I know voxpupli and puppetlabs are working on it. It's why I've been pushing on dropping Puppet 3 after the Foreman 1.15 branching (which is soon).

Klaas-:
@ekohl how to we continue here? it's failing because it pulls https://github.com/puppetlabs/puppetlabs-xinetd master and that one dropped puppet3 support. Should I specify a tag for that repo that works or is it acceptable to fail on the old puppet version?

ehelms:
Hrmm, we might be able to remove xinetd. This puppet module used to wrap foreman_proxy and thus required a lot of the fixtures it had. I'd bet a lot of those are not needed anymore.

ekohl:
The default puppet-foreman_proxy sets up tftp as well. We can set parameters to turn that off as well with the benefit of faster tests.

Klaas-:
@ekohl I fixed the new lint tests and the fixture for puppet3 :)
although I'm not sure why it enforces the ordering arrows to be on the other lines... :)

ekohl:
It's the new style since puppet-lint 2.2.0. We just merged a bunch of the changes for theforeman and we should do the same for katello.

stbenjam:
Ugh, I'd rather disable that lint check, it's hideous.

ehelms:
I agree, its like the leading comma's in JavaScript which were just silly IMO.

PR #116, opened 2017-03-24
Modulesync update

ekohl:


ekohl:
Fails on puppet 3 because puppetlabs-xinetd dropped support for it.

PR #119, opened 2017-04-13
fixes #19269 - allow browsing /pub over https

stbenjam:
Currently the pub template is only in HTTP vhost, but /pub should be available over https too which is in the pulp vhost.


ekohl:
If you rebase this then the tests should pass.

stbenjam:
They do, thanks!

PR #120, opened 2017-04-25
[WIP] Refactor to Puppet 4 types

stbenjam:


metadata.json
```diff
@@ -52,7 +52,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 3.2.0 < 5.0.0"
+      "version_requirement": ">= 4.0.0 < 5.0.0"
```
ekohl: We use 4.6.1 in theforeman and copied that from voxpupuli and puppetlabs. It's considered the first 4 version you can really rely on.

manifests/init.pp
```diff
@@ -5,100 +5,78 @@
...
+  Integer[0, 65535] $qpid_router_hub_port = $foreman_proxy_content::params::qpid_router_hub_port,
+  String $qpid_router_agent_addr = $foreman_proxy_content::params::qpid_router_agent_addr,
+  Integer[0, 65535] $qpid_router_agent_port = $foreman_proxy_content::params::qpid_router_agent_port,
+  String $qpid_router_broker_addr = $foreman_proxy_content::params::qpid_router_broker_addr,
+  Integer[0, 65535] $qpid_router_broker_port = $foreman_proxy_content::params::qpid_router_broker_port,
+  String $qpid_router_logging_level = $foreman_proxy_content::params::qpid_router_logging_level,
+  Stdlib::Absolutepath $qpid_router_logging_path = $foreman_proxy_content::params::qpid_router_logging_path,
+  Boolean $enable_ostree = $foreman_proxy_content::params::enable_ostree,
 ) inherits foreman_proxy_content::params {
   validate_bool($enable_ostree)
```
ekohl: This is redundant now. The 2 other validations as well.

ekohl: The third can be replaced with ```assert_type(String, $pulp_oauth_secret)```. Maybe ```String[5]``` to set a minimum secret length. That would drop the dependency on puppet-common as well.

ekohl:
Merged via https://github.com/Katello/puppet-foreman_proxy_content/issues/122

PR #121, opened 2017-05-02
Fix reverse proxy config

ekohl:
The error code should be 500 according to the document. It looks like a
copy paste error.

The reverse proxy port should be used from the main class so it can be
overridden. Since parent_fqdn is already used from the main class, we
can use the port in the same way.


jlsherrill:
ACK thanks @ekohl !

PR #126, opened 2017-05-11
Correct pulp certificate integration

ekohl:
We need the certs class to be executed for the directories and files
that pulp is using. We can also use the actual class parameters rather
than the params.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -102,23 +102,26 @@ class foreman_proxy_content (
 
   class { '::certs::foreman_proxy':
     hostname => $foreman_proxy_fqdn,
-    require  => Package['foreman-proxy'],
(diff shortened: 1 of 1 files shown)


ekohl:
The issue I'm seeing in pipeline builds is that ```/etc/pki/katello/certs``` does not exist causing ```pulp-gen-ca-certificates``` to fail. That command should not run at all since ```certs::ca``` creates the actual certificates. Somehow this patch isn't working.

ekohl:
I suspect it's because pulp classes are not properly contained so I created https://github.com/Katello/puppet-pulp/pull/239. Now I have another error where Apache can't start because the CA file doesn't exist or is empty.

ekohl:
Squashed the WIP commit, no real content change.

PR #127, opened 2017-06-06
Add crane data dir

jlsherrill:


jlsherrill:
specifying crane_data_dir as part of the pulp class does not work when crane is specified on its own.  I think it makes sense to specify crane on its own here to not have to duplicate all the crane options across the katello, katello_devel, and this module

PR #131, opened 2017-07-26
msync: Puppet 5, parallel tests, .erb templates, cleanups, facter fix

ekohl:


manifests/init.pp
```diff
@@ -217,6 +217,7 @@
     # We can't pull the certs out to the top level, because of how it gets the default
     # parameter values from the main ::certs class.  Kafo can't handle that case, so
     # it remains here for now.
+    include ::puppet
```
ekohl: I'm not 100% sure this is the correct solution, but since we notify it, it must be included. The other solution would be to add a collector but I think that's rather ugly.

PR #132, opened 2017-08-10
seperate parent_fqdn and foreman_fqdn

Klaas-:
also cleanup a few var names to reflect what they actually mean

this is meant as a first step to integrate a pulp master into this module

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -4,7 +4,7 @@
 #
 # === Parameters:
 #
-# $parent_fqdn::                        FQDN of the parent node.
+# $parent_fqdn::                        FQDN of the parent pulp node.
 #
 # $enable_ostree::                      Boolean to enable ostree plugin. This requires existence of an ostree install.
 #
@@ -82,16 +82,16 @@ class foreman_proxy_content (
   include ::foreman_proxy
   include ::foreman_proxy::plugin::pulp
 
-  $pulp = $::foreman_proxy::plugin::pulp::pulpnode_enabled
-  if $pulp {
+  $pulp_node = $::foreman_proxy::plugin::pulp::pulpnode_enabled
+  if $pulp_node {
     assert_type(String[1], $pulp_oauth_secret)
   }
 
   $foreman_proxy_fqdn = $::fqdn
-  $foreman_url = "https://${parent_fqdn}"
-  $reverse_proxy_real = $pulp or $reverse_proxy
+  $foreman_url = $::foreman_proxy::foreman_base_url
+  $setup_reverse_proxy = $pulp_node or $reverse_proxy
 
-  $rhsm_port = $reverse_proxy_real ? {
+  $rhsm_port = $setup_reverse_proxy ? {
     true  => $reverse_proxy_port,
     false => '443'
   }
@@ -112,7 +112,7 @@ class foreman_proxy_content (
     require        => Class['certs'],
   }
 
-  if $pulp or $reverse_proxy_real {
+  if $setup_reverse_proxy {
     class { '::certs::apache':
       hostname => $foreman_proxy_fqdn,
       require  => Class['certs'],
@@ -125,7 +125,7 @@ class foreman_proxy_content (
     }
   }
 
-  if $pulp_master or $pulp {
+  if $pulp_master or $pulp_node {
     if $qpid_router {
       class { '::foreman_proxy_content::dispatch_router':
         require => Class['pulp'],
@@ -141,7 +141,7 @@ class foreman_proxy_content (
     }
   }
 
-  if $pulp {
+  if $pulp_node {
     include ::apache
     $apache_version = $::apache::apache_version
 
--- a/manifests/reverse_proxy.pp
+++ b/manifests/reverse_proxy.pp
@@ -1,7 +1,7 @@
 # Adds http reverse-proxy to parent conf
 class foreman_proxy_content::reverse_proxy (
   $path = '/',
-  $url  = "https://${foreman_proxy_content::parent_fqdn}/",
+  $url  = "${foreman_proxy_content::foreman_url}/",
   $port = $foreman_proxy_content::reverse_proxy_port,
 ) {
   include ::apache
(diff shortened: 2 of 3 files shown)


manifests/init.pp
```diff
@@ -12,6 +12,8 @@
 #
 # === Advanced parameters:
 #
+# $foreman_fqdn::                       FQDN of the foreman server (usually equal to parent_fqdn)
```
ekohl: Maybe parent_fqdn needs a longer description or at least mention that it's the parent pulp node

manifests/init.pp
```diff
@@ -52,6 +54,7 @@
 #
 class foreman_proxy_content (
   String[1] $parent_fqdn = $foreman_proxy_content::params::parent_fqdn,
+  String[1] $foreman_fqdn = $foreman_proxy_content::params::foreman_fqdn,
```
ekohl: ```foreman_proxy``` has a parameter ```foreman_base_url``` and I'm wondering if it should be passed to it, but that would mean changing the ```include``` to a ```class``` and that might break some things.

timogoebel: It will definitely break the installer, as the scenarios include foreman_proxy with parameters.

Klaas-: I think this could be part of a later change if we figure it out :) this is just an initial step to clean up this module and get pulp master setup into this

ekohl: This is already a problem in the current code and this isn't making it much worse.

ekohl: Another thought: instead of exposing the parameter use ```$::foreman_proxy::foreman_base_url```. That way you automatically get a working setup.

templates/_pulp_gpg_proxy.erb
```diff
@@ -1,5 +1,5 @@
-  ProxyPass /katello/api/repositories/ https://<%= @parent_fqdn %>/katello/api/repositories/
+  ProxyPass /katello/api/repositories/ https://<%= @foreman_fqdn %>/katello/api/repositories/
```
ekohl: I think this could be ```<%= @foreman_url %>/katello/api/repositories/```

templates/_pulp_gpg_proxy.erb
```diff
@@ -1,5 +1,5 @@
...
   <Location /katello/api/repositories/>
-    ProxyPassReverse https://<%= @parent_fqdn %>/
+    ProxyPassReverse https://<%= @foreman_fqdn %>/
```
ekohl: I think this could be ```<%= @foreman_url %>/```

manifests/reverse_proxy.pp
```diff
@@ -1,7 +1,7 @@
 # Adds http reverse-proxy to parent conf
 class foreman_proxy_content::reverse_proxy (
   $path = '/',
-  $url  = "https://${foreman_proxy_content::parent_fqdn}/",
+  $url  = "https://${foreman_proxy_content::foreman_fqdn}/",
```
ekohl: This could be ```"${::foreman_proxy_content::foreman_url}/"```

Klaas-:
@ekohl done

manifests/init.pp
```diff
@@ -4,7 +4,7 @@
 #
 # === Parameters:
 #
-# $parent_fqdn::                        FQDN of the parent node.
+# $parent_fqdn::                        FQDN of the parent pulp node.
```
ehelms: We really need a better term than Pulp node given that is/was an actual concept that we are no longer using.

manifests/init.pp
```diff
@@ -82,16 +82,16 @@
...
+  $pulp_node = $::foreman_proxy::plugin::pulp::pulpnode_enabled
+  if $pulp_node {
     assert_type(String[1], $pulp_oauth_secret)
   }
 
   $foreman_proxy_fqdn = $::fqdn
-  $foreman_url = "https://${parent_fqdn}"
-  $reverse_proxy_real = $pulp or $reverse_proxy
+  $foreman_url = $::foreman_proxy::foreman_base_url
+  $setup_reverse_proxy = $pulp_node or $reverse_proxy
```
ehelms: Can you explain cases where the `or` will come into play? Where you'd have one without the other and vice versa?

Klaas-: thats the way it was before :) I think someone had in mind to install a reverse proxy without pulp so thats the way I kept it

manifests/init.pp
Klaas-: what would you suggest eric?

ekohl: I further looked into this: there is no pulp parent, just the Foreman URL and the qpid hub. We can take the former from the proxy config. Not sure what to name the latter.

ehelms: Ideas:

 1) Parent pulp and child pulp
 2) Master pulp and pulp
 3) Parent pulp and pulp
 4) Master pulp and replica pulp

ekohl: It's only used in a pulp node situation to point to the qpid hub. Would ```qpid_hub_fqdn``` make sense?

ehelms: +1 but a slight change to `qpid_router_hub_fqdn` or maybe `qpid_router_hub_addr` to match the param? Then again, looking at (https://github.com/Katello/puppet-foreman_proxy_content/blob/c53398bb2c0e697c6921fcc2f78c23b6b99a81be/manifests/dispatch_router.pp#L45) this seems to already exist and could just be used?

ekohl: That's currently used in the pulp master situation to listen on. I'm not sure if we should combine two opposite behaviors into one parameter.

ehelms: Isn't that the point of configuration? The parameter has the same meaning in both situations, what address should the hub be connecting/talking/listening on? In one case, you want to configure it for the server itself, in another you want it to be pointed at another. If we want to two params, I think we gotta make it clear that is the qpid routers 'parent' or 'master' you want to specify.

ehelms:
@Klaas- Any updated thoughts here? Looks like this will also need a rebase.

ekohl:
Closing this out. I've started https://github.com/theforeman/puppet-foreman_proxy_content/pull/218.

PR #133, opened 2017-08-10
[WIP] pulp master setup via foreman_proxy_content

Klaas-:
this includes #132 
it is supposed to be the second step towards installing pulp masters via foreman_proxy_content. This should work for pulp master (with katello application and without) and pulp nodes.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -16,6 +16,8 @@
 #
 # $pulp_master::                        Whether the foreman_proxy_content should be identified as a pulp master server
 #
+# $pulp_master_standalone::             Whether the pulp master should manage httpd, only useable when having a pulp master without katello on the same server
+#
 # $pulp_admin_password::                Password for the Pulp admin user. It should be left blank so that a random password is generated
 #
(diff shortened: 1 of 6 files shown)


manifests/init.pp
```diff
@@ -162,49 +195,88 @@
...
+      $node_oauth_secret         = undef
+      $node_server_ca_cert       = undef
+    }
+
+
+
+    class { '::pulp':
       broker_url                => "qpid://${qpid_router_broker_addr}:${qpid_router_broker_port}",
       broker_use_ssl            => true,
+      ca_cert                   => $ca_cert,
```
timogoebel: This is currently unset for the "pulp master", see https://github.com/Katello/puppet-katello/blob/master/manifests/init.pp#L167-L197.
We need a setting to set this to `undef` on a standalone pulp master. This is currently not possible.

ekohl:
This is still a good idea, but I'm going to close it due to its age now. We're planning improvements around Pulp deployment and these ideas will be used.

PR #136, opened 2017-08-22
[WIP] Fixes #20852 - Refactor module

ekohl:
This is not ready for review at all and only meant for me to run the forklift pipeline with it.


timogoebel:
@ekohl: I know that is is not meant for a review. But I revied it anyways. This looks already pretty good to me and I'd love to have this as a base for further improvement. What do you think is missing here?

ekohl:
@timogoebel I don't recall if this current iteration actually worked. Other than that I want to rename the ```parent_fqdn``` parameter. It will be a breaking change but in this patch it should be ```qdrouter_hostname``` or similar I think.

timogoebel:
@ekohl: Maybe we can extract that change first to make this less complex. I'm working on a PR to separate the upstreams for qpid and the reverse proxy. I'll send that over in a bit. I'd like if this PR would be just refactoring without changing anything.

ekohl:
@timogoebel that's always a good idea.

ekohl:
@timogoebel rebased. It's not split into a bit more separate commits. I still feel like the pulp refactoring could be split up further for easier review but I think the first commits can be merged already.

manifests/init.pp
```diff
@@ -109,136 +109,47 @@
   Boolean $manage_broker = $foreman_proxy_content::params::manage_broker,
 ) inherits foreman_proxy_content::params {
   include ::certs
+  include ::certs::foreman_proxy
   include ::foreman_proxy
   include ::foreman_proxy::plugin::pulp
+  Class['certs::foreman_proxy'] ~> Class['foreman_proxy::service']
 
-  $pulp = $::foreman_proxy::plugin::pulp::pulpnode_enabled
+  $pulp_node = $::foreman_proxy::plugin::pulp::pulpnode_enabled
```
ehelms: We really should replace the node nomenclature.

manifests/pulp/master.pp
```diff
@@ -0,0 +1,65 @@
...
+# usually via puppet-katello. Here we only configure the qpid router (if
+# desired) and crane.
+class foreman_proxy_content::pulp::master (
+  Boolean $qpid_router = $::foreman_proxy_content::qpid_router,
+  String $hub_host = $::foreman_proxy_content::qpid_router_hub_addr,
+  Integer[0, 65535] $hub_port = $::foreman_proxy_content::qpid_router_hub_port,
+  String $broker_host = $::foreman_proxy_content::qpid_router_broker_addr,
+  Integer[0, 65535] $broker_port = $::foreman_proxy_content::qpid_router_broker_port,
+) {
+  if $qpid_router {
```
ehelms: Maybe this should be it's own manifest? Pulp does not actually talk to the router and you may choose to deploy the router separately from your Pulp host or even your qpid host. That is, you could have Pulp server, Qpid server and Qdrouter running all separately or router and qpid paired.

manifests/pulp/master.pp
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  include ::certs::apache
+  class { '::pulp::crane':
+    cert      => $::certs::apache::apache_cert,
+    key       => $::certs::apache::apache_key,
+    ca_cert   => $::certs::ca_cert,
+    data_dir  => '/var/lib/pulp/published/docker/v2/app',
+    subscribe => Class['certs::apache'],
+  }
```
ehelms: This can run on a Smart proxy as well serving up docker content off the Pulp server there.

manifests/pulp/node.pp
```diff
@@ -0,0 +1,105 @@
+# The pulp node configuration
+#
+# In this scenario we assume there is another pulp instance somewhere and this
+# talks to a qpid router somewhere.
```
ehelms: This is technically speaking incorrect. This [connection diagram](https://raw.githubusercontent.com/ehelms/connection_diagram/master/katello.png) may help with wrapping your brain around all the connection points. 

The dispatch router serves as a public interface for services that need to talk to qpid. Generally speaking thats only clients talking to dispatch router which then talks to either Qpid directly (on a server) or talks to another dispatch router (on a smart proxy). Currently Pulp talks directly to qpid over localhost. If we worked on a spread deployment and split out qpid we'd need to consider having Pulp talk to the router but for current setups it does not.

manifests/pulp/node.pp
```diff
@@ -0,0 +1,105 @@
...
+  include ::apache::mod::proxy
+  include ::apache::mod::proxy_http
+
+  file {'/etc/httpd/conf.d/pulp_nodes.conf':
+    ensure  => file,
+    content => template('foreman_proxy_content/pulp_nodes.conf.erb'),
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0644',
+  }
```
ehelms: This block can be dropped from sight including the template.

manifests/pulp/node.pp
```diff
@@ -0,0 +1,105 @@
...
+    ssl_cert_db            => $::certs::nss_db_dir,
+    ssl_cert_password_file => $::certs::qpid::nss_db_password_file,
+    ssl_cert_name          => 'broker',
+    interface              => 'lo',
+    subscribe              => Class['certs', 'certs::qpid'],
+  }
+
+  include ::certs::apache
+  include ::certs::qpid_client
+  class { '::pulp':
```
ehelms: Is it worth considering centralizing Pulp install and configuration in one place at this juncture? I don't want to over complicate this large re-factor but as we've pointed out in the past we setup Pulp here and in puppet-katello.

ekohl: I considered moving it all into puppet-katello, but you're right we have overlap. We should at least centralize pulp in one place.

manifests/init.pp
ekohl: Indeed, but what should be the correct name?

manifests/pulp/master.pp
ekohl: I think you're right.

manifests/init.pp
ehelms: I reached out to pulp-dev list to look for commonly accepted nomenclature. The best I could arrive at is 'parent' and 'child' from the responses I got.

ekohl: Given in how many places this naming is currently used, I'm leaning to keeping that outside of this refactor.

manifests/pulp/master.pp
ekohl: That's why I added ```enable_crane``` there.

ekohl:
Closing this now but will pick this up via other PRs.

PR #140, opened 2017-10-08
use foreman_url for reverse proxy upstream

timogoebel:
This is step one of dropping `parent_fqdn` parameter. This takes the upstream URL for the reverse proxy from `foreman_proxy` module.
Step two will be to rename `parent_fqdn` (after this change: only the upsteam of the qpid router hub) to something like `qpid_broker_fqdn`.


ekohl:
Thanks!

PR #141, opened 2017-10-08
introduce manage broker parameter

timogoebel:
This extracts the management of the qpid message broker for a not-master pulp instance (pulp node) to a separate class. In addition, a parameter was added if a user does not want puppet or this module to manage it.


timogoebel:
Renamed `git mv spec/classes/broker_spec.rb spec/classes/foreman_proxy_content__broker_spec.rb`.

ekohl:
Thanks!

PR #142, opened 2017-10-08
extract pub dir setup

timogoebel:


spec/classes/foreman_proxy_content__pub_dir_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+describe 'foreman_proxy_content::pub_dir' do
+  on_os_under_test.each do |os, facts|
+    context "on #{os}" do
+      let :facts do
+        facts
+      end
+
+      it { is_expected.to compile.with_all_deps }
+      it { should contain_package('katello-client-bootstrap') }
```
ekohl: Let's keep ```is_expected.to``` vs ```should``` consistent per file.

ekohl:
Thanks!

PR #148, opened 2017-10-27
introduce rhsm_hostname parameter

timogoebel:
This adds a parameter `rhsm_hostname` that allows overwriting the hostname that is configured into the client cert rpm.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -30,7 +30,9 @@
 #
 # $reverse_proxy_port::                 Reverse proxy listening port
 #
-# $rhsm_url::                           The URL that the RHSM API is rooted at
+# $rhsm_hostname::                      The hostname that the RHSM API is rooted at
+#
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -30,7 +30,9 @@
 #
 # $reverse_proxy_port::                 Reverse proxy listening port
 #
-# $rhsm_url::                           The URL that the RHSM API is rooted at
+# $rhsm_hostname::                      The hostname that the RHSM API is rooted at
+#
+# $rhsm_url::                           The URL path that the RHSM API is rooted at
```
ekohl: Looks like this parameter is badly named. 👍 for at least improving the documentation.

timogoebel: Yeah, the parameter name is very misleading. `rhsm_path` or something would be better.

PR #150, opened 2017-11-06
Update Github URLs

ekohl:


ehelms:
This one appears to have a test failure.

ekohl:
Looks like we still had puppet-crane in there and since it didn't move ... Good riddance :)

PR #151, opened 2017-11-07
Fixes #21464 - Remove Pulp oauth support

sean797:


ehelms:
Small breakage here in the linting. Looks good otherwise.

sean797:
Ah thanks @ehelms. Updated

PR #154, opened 2018-01-03
refs #21350 - allow config of TLS version on 8443

stbenjam:


ehelms:
We may also need to set this on `pulp::crane`.

stbenjam:
@ehelms Updated

ekohl:
> We may also need to set this on pulp::crane.

Since that defaults to undef it can already be overridden on a global apache level.

manifests/init.pp
```diff
@@ -127,6 +131,7 @@
       key      => $certs::apache::apache_key,
       ca_cert  => $certs::ca_cert,
       data_dir => '/var/lib/pulp/published/docker/v2/app',
+      ssl_protocol => $ssl_protocol,
```
ekohl: Alignment is off

stbenjam: Fixed

PR #158, opened 2018-01-25
Allow use of puppet-qpid 4.X

ehelms:


metadata.json
```diff
@@ -30,7 +30,7 @@
     },
     {
       "name": "katello-qpid",
-      "version_requirement": ">= 3.0.0 < 4.0.0"
+      "version_requirement": ">= 3.0.0 < 5.0.0"
```
ekohl: It will actually not work with < 4.0 versions

ehelms:
Good catch -- updated

ekohl:
https://github.com/theforeman/puppet-foreman_proxy/pull/407 will release 7.0 and would break metadata again. Mind including that too? :)

PR #160, opened 2018-02-26
make all repo types configurable

jlsherrill:


manifests/init.pp
```diff
@@ -190,10 +203,11 @@
       require => Class['certs'],
     }
     ~> class { '::pulp':
-      enable_rpm             => true,
-      enable_puppet          => true,
-      enable_docker          => true,
       enable_ostree          => $enable_ostree,
+      enable_rpm             => $enable_yum or $enable_file,
```
sean797: shouldn't there be `enable_iso => $enable_file` ?

It was added recently ;) https://github.com/theforeman/puppet-pulp/commit/1faa3e7d204edd828031627bca494ae1c654388d

sean797: I think your also need to update the minimum version in `metadata.json`

jlsherrill: ahhhh, didn't realize this was a thing now.  @sean797  It doesn't look like that actually would work on its own without enabling yum (since the yum plugin wouldn't be installed, or am i misreading something?)

sean797: Ah yes, I think you'ee right :( https://github.com/theforeman/puppet-pulp/pull/309 :wink:

jlsherrill: updated!

jlsherrill:
ahhh right, updated!

jlsherrill:
@sean797 look good now?

PR #166, opened 2018-06-07
refs #21350 - use ssl configuration for server profile

stbenjam:
This was on the wrong ssl-profile, the server profile needs it, not client.


ehelms:
@stbenjam Could you do an updated release of the module?

PR #168, opened 2018-06-20
Only set up puppet integration if needed

ekohl:
This still isn't quite perfect, but right now I can't think of a better way to correctly determine if puppet integration is actually needed.  Ideally we would get rid of the $puppet parameter but then we need to resort to defined('puppet') and that's compile order dependent.

This does mean at worst we pull in the client and not the server.

We can also change the notify to a before because the integration is just scripts that read files every run. There is no daemon that caches it.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -259,11 +259,12 @@ class foreman_proxy_content (
     # parameter values from the main ::certs class.  Kafo can't handle that case, so
     # it remains here for now.
     include ::puppet
-    include ::puppet::server
-    class { '::certs::puppet':
-      hostname => $foreman_proxy_fqdn,
-      require  => Class['certs'],
-      notify   => Class['puppet'],
+    if $::puppet::server and $::puppet::server::foreman {
+      class { '::certs::puppet':
+        hostname => $foreman_proxy_fqdn,
+        require  => Class['certs'],
+        before   => Class['puppet'],
+      }
     }
   }
 


manifests/init.pp
```diff
@@ -259,11 +259,12 @@
     # parameter values from the main ::certs class.  Kafo can't handle that case, so
     # it remains here for now.
     include ::puppet
-    include ::puppet::server
```
sean797: Why remove this?

ekohl: This was silently pulling in the puppet server which users might not want. The main puppet class has a parameter whether to include it and also correct chaining. You might be right that there are cases where a user has manually included it and this isn't correct. Coupling to other modules is annoying.

ekohl: @sean797 do you think I should add a `fail()` instead when those variables are false?

ekohl: What do you man by `it` here? The include or a `fail()` message?

sean797: sorry the include, so

```
if $::puppet::server and $::puppet::server::foreman {
  include ::puppet::server
```

ekohl: https://github.com/theforeman/puppet-puppet/blob/837e65929a7ba12e2bc730ec7e817c33acda3f57/manifests/init.pp#L743-L746 should already take care of that

sean797: in that case, I'm okay with this :+1:

PR #180, opened 2018-10-23
Release 7.3.1

ehelms:


ehelms:
Fair point, I was trying me best to follow the docs. I think we can make that nugget clearer.

ekohl:
Yes, it was documenting my steps but clearly I did some additional ones without thinking.

PR #185, opened 2018-11-26
Update qpid client certificate

ehelms:


manifests/init.pp
```diff
@@ -220,8 +220,8 @@
       default_password       => $pulp_admin_password,
       messaging_transport    => 'qpid',
       messaging_auth_enabled => false,
-      messaging_ca_cert      => pick($pulp_ca_cert, $certs::ca_cert),
-      messaging_client_cert  => $certs::messaging_client_cert,
+      messaging_ca_cert      => pick($pulp_ca_cert, $::certs::ca_cert),
```
evgeni: I thought "top scope" `$::` was deprecated in recent Puppets? @ekohl?

ekohl: It's no longer needed so it's more and more becoming best practice to leave them out. I've started to convert our modules and this one has also been converted.

ehelms:
Updated

PR #189, opened 2019-01-09
Allow docker pull on $HOSTNAME

ehelms:
This is to allow docker pull against $HOSTNAME without having to
specify port in the registry declaration.


ehelms:
Related -- https://github.com/theforeman/puppet-katello/pull/276

I was not able to find a spot to do this all here as the handling of Apache is just different enough on a server vs. stand alone smart proxy.

templates/redirect_crane.erb
```diff
@@ -0,0 +1,2 @@
+Redirect /v1 https://<%= scope['fqdn'] %>:5000/v1
+Redirect /v2 https://<%= scope['fqdn'] %>:5000/v2
```
evgeni: same comment as in https://github.com/theforeman/puppet-katello/pull/276#pullrequestreview-190796622

ehelms:
@evgeni Tested and worked for me with rewrite update, please review

ehelms:
See related puppet-katello PR

PR #194, opened 2019-02-19
Fixes #26088 - don't set ssl_proxy_machine_cert for rhsm proxy

iNecas:
It seems SSLProxyMachineCertificateFile option supports only pkcs1
format (containing BEGIN RSA PRIVATE KEY), while it's not possible to use
this kind of keys in FIPS mode. The implication of this is we can't use
this directive in FIPS enabled environment.

Additionally, I don't see any reason to use this directive in the first
place so it seems it should be safe to remove. Please speak up with if you can think of one.

What I've tried is to comment out this setting and doing this from registered host worked just fine:

curl --cert /etc/pki/consumer/cert.pem --key /etc/pki/consumer/key.pem -k https://mycapsule.example.com:8443/rhsm/consumers/$(subscription-manager identity | head -n1 | grep -o '\S*$')

### DIFF
--- a/manifests/reverse_proxy.pp
+++ b/manifests/reverse_proxy.pp
@@ -20,7 +20,6 @@ class foreman_proxy_content::reverse_proxy (
     ssl                    => true,
     ssl_proxyengine        => true,
     ssl_proxy_ca_cert      => $certs::ca_cert,
-    ssl_proxy_machine_cert => $certs::foreman_proxy::foreman_proxy_ssl_client_bundle,
     ssl_cert               => $certs::apache::apache_cert,
     ssl_key                => $certs::apache::apache_key,
     ssl_chain              => $certs::katello_server_ca_cert,
--- a/manifests/reverse_proxy.pp
+++ b/manifests/reverse_proxy.pp
@@ -12,30 +12,30 @@ class foreman_proxy_content::reverse_proxy (
   Class['certs', 'certs::ca', 'certs::apache', 'certs::foreman_proxy'] ~> Class['apache::service']
 
   apache::vhost { 'katello-reverse-proxy':
-    servername             => $foreman_proxy_content::foreman_proxy_fqdn,
-    port                   => $port,
-    docroot                => '/var/www/',
-    priority               => '28',
-    ssl_options            => ['+StdEnvVars', '+ExportCertData', '+FakeBasicAuth'],
-    ssl                    => true,
-    ssl_proxyengine        => true,
-    ssl_proxy_ca_cert      => $certs::ca_cert,
-    ssl_cert               => $certs::apache::apache_cert,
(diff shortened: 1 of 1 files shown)


iNecas:
Fixing the CI issues in separate commit, just to show the required change vs. the whitespace updates

iNecas:
Updated the commit messages to match the original redmine issue

ehelms:
How extensively have you tested this? I'm fairly certain, reaching way back in my memory, that this exists to ensure mutual authentication between the Proxy and the main server, and to ensure that client SSL headers are properly passed and re-encoded when passing through the proxy to the server Apache and eventually to the Ruby application.

ehelms:
If you edit the keys header to `BEGIN RSA PRIVATE KEY` and the footer subsequently, httpd seemed to start. Does this 'break' FIPS?

ehelms:
I am seeing the following fill the logs on every request:

```
[Tue Feb 19 22:05:35.517737 2019] [ssl:warn] [pid 20054] AH02268: Proxy client certificate callback: (pipeline-proxy-nightly-centos7.example.com:443) downstream server wanted client certificate but none are configured
```

ekohl:
> It seems SSLProxyMachineCertificateFile option supports only pkcs1 format

The docs say it must be PEM encoded so perhaps we're generating incorrect PEM files?

https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#SSLProxyMachineCertificateFile
> File of concatenated PEM-encoded client certificates and keys to be used by the proxy

iNecas:
The problem is in fips environment, you can't generate the pkcs1 format (see here for exmaple http://openssl.6102.n7.nabble.com/Not-getting-quot-RSA-quot-keyword-for-a-key-in-fips-mode-td58588.html), so the pkcs8 is generated instaed and https://access.redhat.com/solutions/3214421 says "Note: Apache httpd mod_ssl [SSLProxyMachineCertificateFile] does not support PKCS#8 key format.".

The PEM files are correct (the work for other use cases, as proxy authenticating against the Foreman), it's just the mod_ssl refusing to use this format, probably for history reason.

@ehelms I was testing it functionality wise, where I saw no difference, not checked the logs though

iNecas:
Surprisingly, just changing the `BEGIN PRIVATE KEY` to `BEGIN RSA PRIVATE KEY` without actually changing the content of the key works in FIPS.

I'm thinking of adding `force_rsa` option for the key_bundle and other resources to enforce this format for places we know require it.

iNecas:
Here is the alternative path that keeps the setting there but makes sure the RSA word is still around https://github.com/theforeman/puppet-certs/pull/243

iNecas:
Closing in favor of https://github.com/theforeman/puppet-certs/pull/243

PR #207, opened 2019-07-09
Allow customization of pub dir on foreman proxy server

snagoor:
We have a hardcoded value `Options` for `Location /pub` on the foreman proxy node 

https://github.com/theforeman/puppet-foreman_proxy_content/blob/7c479d473cd4e4a54483e9495bf73c466cfce2b7/templates/httpd_pub.erb#L7

We should allow over-riding this value through hiera because some users don't like to expose `/pub` per security guidelines.

~~~
# vi /etc/foreman-installer/custom-hiera.yaml
foreman_proxy_content::pub_dir::pub_dir_options: "+FollowSymLinks"
~~~

### DIFF
--- a/manifests/pub_dir.pp
+++ b/manifests/pub_dir.pp
@@ -3,6 +3,7 @@
 # for download
 class foreman_proxy_content::pub_dir (
   String $servername = $facts['fqdn'],
+  String $pub_dir_options = '+FollowSymLinks +Indexes',
 ) {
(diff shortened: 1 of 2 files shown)


ekohl:
Thanks! I've taken the liberty to open a Redmine issue for this and modified the commit message to close that. This allows us to properly track it and close a Bugzilla using automation.

PR #210, opened 2019-08-22
Fixes #27687 - Add vhost_params and proxy_pass_params

ekohl:
This allows customizing the Apache vhost and ProxyPass via hiera giving more flexibility. Also does some more cleaning up as preparation.

### DIFF
--- a/manifests/reverse_proxy.pp
+++ b/manifests/reverse_proxy.pp
@@ -10,12 +10,18 @@
 #   The port to listen on
 # @param ssl_protocol
 #   The ssl protocol(s) to accept
+# @param vhost_params
+#   Any parameters to pass to the apache::vhost resource
+# @param proxy_pass_params
+#   Any parameters to pass to the proxy_pass param of the apache::vhost resource
 class foreman_proxy_content::reverse_proxy (
   Stdlib::Unixpath $path = '/',
   Stdlib::Httpurl $url = "${foreman_proxy_content::foreman_url}/",
   Stdlib::Fqdn $servername = $foreman_proxy_content::foreman_proxy_fqdn,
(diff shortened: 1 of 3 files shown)


manifests/reverse_proxy.pp
```diff
@@ -1,9 +1,27 @@
...
+# @param servername
+#   The Apache vhost server name to listen on
+# @param port
+#   The port to listen on
+# @param ssl_protocol
+#   The ssl protocol(s) to accept
+# @param vhost_params
+#   Any parameters to pass to the apache::vhost resource
+# @param proxy_pass_params
+#   Any parameters to pass to the apache::vhost resource
```
evgeni:
```suggestion
#   Any parameters to pass to the proxy_pass param of the apache::vhost resource
```

manifests/reverse_proxy.pp
```diff
@@ -1,9 +1,27 @@
...
 class foreman_proxy_content::reverse_proxy (
-  $path = '/',
-  $url  = "${foreman_proxy_content::foreman_url}/",
-  $port = $foreman_proxy_content::reverse_proxy_port,
-  $ssl_protocol = undef,
+  String $path = '/',
```
evgeni: Is there a better type to denote "an absolute (looking) path"?

ekohl: There is `Stdlib::Unixpath` which is strictly speaking not a URL path, but might be good enough.

evgeni: Yeah, I thought about Stdlib::Absolutepath, but yours is better.

ekohl:
Updated

PR #215, opened 2019-11-22
Move the certs::apache definition into reverse_proxy

ekohl:
It was already included and internally already had the correct triggers. This simplifies the code.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -165,7 +165,7 @@ class foreman_proxy_content (
     require        => Class['certs'],
   }
 
-  if $pulp or $reverse_proxy_real {
+  if $reverse_proxy_real {
     class { 'foreman_proxy_content::reverse_proxy':
       path         => '/',
       url          => "${foreman_url}/",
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -166,16 +166,11 @@
   }
 
   if $pulp or $reverse_proxy_real {
-    class { 'certs::apache':
-      hostname => $foreman_proxy_fqdn,
-      require  => Class['certs'],
-    }
-    ~> class { 'foreman_proxy_content::reverse_proxy':
```
ehelms: Does moving this still trigger refreshes if the certificates change? e.g. restarting Apache

ekohl: I think this should be sufficient:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/a622bd3fb2f5a8eb6b1881d1828a51dd0f40800c/manifests/reverse_proxy.pp#L30

ekohl:
Actually, looking closer this is used in multiple places.

PR #217, opened 2019-11-22
Derive $pulp_master variable

ekohl:
We can know whether we're a pulp master depending on the proxy setting. There's no need to duplicate it.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -165,17 +165,18 @@ class foreman_proxy_content (
     require        => Class['certs'],
   }
 
-  if $pulp or $reverse_proxy_real {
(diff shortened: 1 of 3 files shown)


ehelms:
Will need a corresponding installer migration to remove it ehh?

ekohl:
Correct. It won't break the installer, but it's useless garbage to keep around in the answers file.

ekohl:
Updated to ensure the correct resources are included at the right time. That was probably already an issue before, but didn't surface due to limited testing. https://github.com/theforeman/foreman-installer/pull/410 should be the installer side.

PR #218, opened 2019-11-22
Rewrite Pulp handling

ekohl:
This is a WIP PR that moves all Pulp handling to a central class. It integrates the handling that puppet-katello did in the past.

This builds on a bunch of other PRs:
* https://github.com/theforeman/puppet-foreman_proxy_content/pull/132
* https://github.com/theforeman/puppet-foreman_proxy_content/pull/133
* https://github.com/theforeman/puppet-foreman_proxy_content/pull/136
* https://github.com/theforeman/puppet-foreman_proxy_content/pull/217


manifests/pulp.pp
```diff
@@ -0,0 +1,135 @@
...
+    manage_httpd           => $manage_httpd,
+    manage_plugins_httpd   => true,
+    manage_squid           => true,
+    messaging_auth_enabled => false,
+    messaging_ca_cert      => $messaging_client_cert,
+    messaging_client_cert  => $messaging_client_cert,
+    messaging_transport    => 'qpid',
+    messaging_url          => $messaging_url,
+    num_workers            => $foreman_proxy_content::pulp_num_workers,
+    proxy_password         => $foreman_proxy_content::pulp_proxy_password, # TODO: not in katello
```
ekohl: @jlsherrill  I recall you said Katello starts to configure Pulp 2 resources with explicit proxies rather than a server wide default. Does that mean we can drop proxy configs?

Note that previously only the Pulp 2 node had proxies configured where the Pulp 2 master had none. Does a node ever pull something from the internet? My impression was that they only mirror the Pulp 2 master.

ehelms:
> It integrates the handling that puppet-katello did in the past.

Does that imply this drops the Pulp handling in puppet-katello or am I reading past the idea?

ekohl:
> Does that imply this drops the Pulp handling in puppet-katello or am I reading past the idea?

No, that's exactly the goal. https://github.com/theforeman/puppet-katello/pull/303 should reflect that side.

One huge motivator for this was also to figure out how we build vhosts. Currently the pulp vhost is used. Given we'll need to do more, I'm leaning to dropping that and explicitly defining a vhost in this module. We already have the reverse proxy on port 8443. That could actually be moved to port 443 as well (though at a later stage).

manifests/pulp.pp
```diff
@@ -0,0 +1,201 @@
...
+      enable_crane           => true, # requires https://github.com/theforeman/puppet-pulp/pull/378
+      enable_deb             => $enable_deb,
+      enable_docker          => $enable_docker,
+      enable_iso             => $enable_file,
+      enable_katello         => !$is_mirror,
+      enable_ostree          => $enable_ostree,
+      enable_puppet          => $enable_puppet,
+      enable_rpm             => $enable_yum,
+      manage_db              => $mongodb_manage_db,
+      manage_httpd           => $is_mirror,
```
wbclark: I assume `$is_mirror` has value `true` for the proxy scenario and `false` for katello. Can I take the same approach when building the logic for pulp3?

ekohl: It is sort of based on Pulp 3 where there's no longer the concept of a master and a node. Just the `is_mirror` parameter. I've started to like that.

evgeni:
Ewoud says NEIN

PR #220, opened 2019-11-29
Use Stdlib::Port where appropriate

ekohl:


evgeni:
this is only present in stdlib 4.25 and newer (https://github.com/puppetlabs/puppetlabs-stdlib/commit/da9a48d7504612b68cfa0fc69ea4506893f1dad9), so we should bump our requirements, ACK otherwise

ekohl:
Updated to require 4.25.0

PR #221, opened 2019-11-29
Make dispatch router standalone classes

ekohl:
This makes it clearer how it's used and allows standalone inclusion.

### DIFF
--- a/manifests/dispatch_router.pp
+++ b/manifests/dispatch_router.pp
@@ -1,14 +1,41 @@
-# == Class: foreman_proxy_content::dispatch_router
+# @summary Install and configure Qpid Dispatch Router
 #
-# Install and configure Qpid Dispatch Router
+# @param agent_addr
+#   Listener address for goferd agents
 #
+# @param agent_port
+#   Listener port for goferd agents
+#
+# @param logging_level
+#   Logging level of dispatch router (e.g. info+ or debug+)
+#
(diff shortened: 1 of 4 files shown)


ekohl:
My main question is whether the `connector` and `hub` should have parameters of `dispatch_router` so they're complete classes.

ehelms:
Can you expand on the question you posed? I think this is a bit of puppetese I don't understand.

ekohl:
The current intended use is (as a connector):
```puppet
class { 'foreman_proxy_content::dispatch_router': 
  agent_addr => $addr,
  # ...
}
class { 'foreman_proxy_content::dispatch_router::connector':
  # connector options
}
```

I'm wondering if it should only call connector and connector should pass options to `foreman_proxy_content::dispatch_router`.

At this point I think I'd start with the current implementation but would be open to other suggestions.

ehelms:
If the base dispatch_router class is more of a common setup for both connector and hub, I would just have users call connector and hide the router part. If ther eare enough configurables on the dispatch_router itself, then it makes sense to me to keep them separate.

ekohl:
The `connector` is what we deploy on a content proxy and the `hub` is what we deploy on a katello scenario. It would duplicate parameters but the calling code is more straight forward.

ehelms:
Sounds good to me

PR #222, opened 2019-12-03
Add initial pulp3 support

wbclark:
This is a Work In Progress pull request to get feedback on adding pulp3 support. I believe the template content is not correct for the foreman-proxy-content scenario at this time.


templates/pulpcore-apache.conf.erb
```diff
@@ -0,0 +1,30 @@
+<VirtualHost *:80>
```
ehelms: Note that we can't do a full vhost here, Foreman on a server defines the vhost and we are dropping a 'fragment' into a directory that gets included into the vhost definition for us. We only really want the proxy pass definitions further down in this template.

manifests/reverse_proxy.pp
```diff
@@ -67,4 +67,10 @@
     ],
     *                      => $vhost_params,
   }
+
+  if $foreman_proxy_content::enable_pulp3 {
```
ehelms: I would move this into `init.pp` for now and also add `and pulp_master` to the conditional to only do this on a server.

manifests/init.pp
```diff
@@ -99,6 +99,9 @@
   Optional[Stdlib::Absolutepath] $pulp_ca_cert = $foreman_proxy_content::params::pulp_ca_cert,
   Integer[0] $pulp_worker_timeout = $foreman_proxy_content::params::pulp_worker_timeout,
 
+  Boolean $pulp3 = $foreman_proxy_content::params::pulp3,
+  Boolean $pulp_master = $foreman_proxy_content::params::pulp_master,
```
ehelms: This is already defined -- https://github.com/theforeman/puppet-foreman_proxy_content/pull/222/files#diff-60ae41fd0a31977447947f59940ee9a4R142

manifests/init.pp
```diff
@@ -254,6 +258,12 @@
     }
   }
 
+  if $pulp3 and $pulp_master {
+    foreman::config::apache::fragment { 'pulp':
```
ehelms: For now, you could put the pulpcore deinfition here as well:

```
class { 'pulpcore':
  manage_apache: false
}
```

And add it as a dependency

wbclark: @ehelms done, thanks

templates/pulpcore-apache.conf.erb
```diff
@@ -0,0 +1,6 @@
+  ProxyPass <%= scope['pulpcore::apache::api_path'] %> <%= scope['pulpcore::apache::api_url'] %>
+  ProxyPassReverse <%= scope['pulpcore::apache::api_path'] %> <%= scope['pulpcore::apache::api_path'] %>
+  ProxyPassReverse <%= scope['pulpcore::apache::api_path'] %> <%= scope['pulpcore::apache::api_url'] %>
```
ekohl: Is the duplicate definition here intended?

wbclark:
```
# grep -n Proxy /etc/httpd/conf.d/10-pulp.conf 
25:  ## Proxy rules
26:  ProxyRequests Off
27:  ProxyPreserveHost Off
28:  ProxyPass /pulp/api http://127.0.0.1:24817/pulp/api
29:  ProxyPassReverse /pulp/api /pulp/api
30:  ProxyPassReverse /pulp/api http://127.0.0.1:24817/pulp/api
31:  ProxyPass /pulp/content http://127.0.0.1:24816/pulp/content
32:  ProxyPassReverse /pulp/content /pulp/content
33:  ProxyPassReverse /pulp/content http://127.0.0.1:24816/pulp/content
```
is what is getting created by puppet-pulpcore itself as defined at https://github.com/theforeman/puppet-pulpcore/blob/master/manifests/apache.pp#L16-L27

are the definitions there redundant as well?

ekohl: That's a good question and I think so. Looks to me like it should be `'reverse_urls' => [$api_url],` and probably also incorrect in `reverse_proxy.pp` in this module because I copied it from there.

manifests/init.pp
```diff
@@ -254,6 +258,15 @@
     }
   }
 
+  if $pulp3 and $pulp_master {
+    class { 'pulpcore':
+      manage_apache => false,
+    }
+    foreman::config::apache::fragment { 'pulp':
+      ssl_content => file('foreman_proxy_content/pulpcore-apache.conf.erb'),
```
ehelms: I think we want to use `template` instead of `file` here.

manifests/init.pp
ehelms: This fragment might should be named `pulpcore` given we already have one named `pulp` and they will clash:

```
[ERROR 2019-12-06T02:12:44 verbose]  Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Foreman::Config::Apache::Fragment[pulp] is already declared at (file: /usr/share/foreman-installer/modules/katello/manifests/pulp.pp, line: 39); cannot redeclare (file: /usr/share/foreman-inst
aller/modules/foreman_proxy_content/manifests/init.pp, line: 265) (file: /usr/share/foreman-installer/modules/foreman_proxy_content/manifests/init.pp, line: 265, column: 5)
```

templates/pulpcore-apache.conf.erb
wbclark: I've removed the redunant ProxyPassReverse declaration in all locations of which I am aware.

manifests/init.pp
wbclark: done, thanks

manifests/init.pp
wbclark: done, thanks

ehelms:
I'd be ok driving it through the configuration from foreman_proxy... I think in which case we should introduce a `pulp3_master` variable and ensure it conflicts with `pulp3_mirror`. We could for now also remove the `pulp3_mirror` parameter until we are ready to support it.

Perhaps then in the installer we can have a hook that checks if katello module is not enabled, katello_master cannot be set to true.

wbclark:
@ekohl @ehelms OK. To confirm, for now we will remove the `$pulp3` parameter from FPC and instead define `$pulp3 = $foreman_proxy::plugin::pulp::pulp3_enabled` and `$pulp3_master = $foreman_proxy::plugin::pulp::pulp3_master` (which we will add to puppet-foreman_proxy).

I've opened https://github.com/theforeman/puppet-foreman_proxy/pull/551 for that and will revise this accordingly.

ekohl:
> I'd be ok driving it through the configuration from foreman_proxy... I think in which case we should introduce a `pulp3_master` variable and ensure it conflicts with `pulp3_mirror`. We could for now also remove the `pulp3_mirror` parameter until we are ready to support it.

I disagree. It's a boolean so when the value is false, it's what we consider in Pulp 2 a master. When it's true, in Pulp 2 it would be a node.

metadata.json
```diff
@@ -35,6 +35,10 @@
     {
       "name": "katello/certs",
       "version_requirement": ">= 6.0.0 < 7.0.0"
+    },
+    {
+      "name": "theforeman/pulpcore",
```
ekohl: This isn't released yet so it probably breaks things. I was actually thinking about adding it to `.fixtures.yml` first so you can at least run unit tests.

wbclark: Thanks - I've added `pulpcore`, `postgresql`, and `redis` to `.fixtures.yml`

manifests/init.pp
```diff
@@ -254,6 +256,18 @@
     }
   }
 
+  if $pulp3 and $pulp3_mirror {
+    class { 'pulpcore':
```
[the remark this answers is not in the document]
wbclark: I've opened https://github.com/theforeman/puppet-pulpcore/pull/45 which I believe is necessary to support this from the `pulpcore` side.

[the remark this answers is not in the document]
ekohl: Merged. Thanks! Please align the arguments

manifests/init.pp
```diff
@@ -253,6 +255,19 @@
     }
   }
 
+  if $pulp3 and $pulp3_mirror {
```
ekohl: The logic should be reversed: do nothing on mirror
```suggestion
  if $pulp3 and !$pulp3_mirror {
```

spec/classes/foreman_proxy_content_spec.rb
```diff
@@ -23,14 +23,16 @@
           class { 'foreman_proxy::plugin::pulp':
             enabled          => false,
             pulpnode_enabled => true,
+            pulp3_enabled    => true,
```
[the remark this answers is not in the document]
wbclark: Done (mostly). I'm not sure if I got the `ssl_content` test quite right yet.

ehelms:
We will at some point want to handle serving repositories over HTTP but I think starting with HTTPS is a good start.

metadata.json
ekohl: As discussed, we'll not do a release and keep it as a soft dependency for now. Let foreman-installer pull it in for us.

metadata.json
```diff
@@ -35,6 +35,10 @@
...
+      "version_requirement": ">= 0.0.1 < 0.2.0"
```
ehelms: Since we aren't releasing puppet-pulpcore yet, I think this has to get dropped here

wbclark: got it, thanks

PR #225, opened 2019-12-16
Install file and container plugins for Pulp 3

ehelms:


manifests/init.pp
```diff
@@ -263,6 +263,9 @@
       servername    => $foreman::servername,
     }
 
+    include pulpcore::plugin::file
```
ekohl: Could you respect the flags?
```suggestion
    if $enable_file {
      include pulpcore::plugin::file
    }
```

manifests/init.pp
```diff
@@ -263,6 +263,9 @@
...
+    include pulpcore::plugin::container
```
ekohl: We should also rename `$enable_docker`, but let's just reuse it now.
```suggestion
    if $enable_docker {
      include pulpcore::plugin::container
    }
```

spec/classes/foreman_proxy_content_spec.rb
```diff
@@ -61,7 +61,7 @@ class { 'foreman_proxy::plugin::pulp':
 
         it do
           is_expected.to contain_foreman__config__apache__fragment('pulpcore')
-		  .with_ssl_content(%r{ProxyPass /pulp/api http://127\.0\.0\.1:24817/pulp/api})
+		  .with_ssl_content(%r{ProxyPass /pulp/api/v3 http://127\.0\.0\.1:24817/pulp/api/v3})
```
wbclark: Nice catch, thanks

manifests/init.pp
ehelms: Sure, I often forget about those things since we deploy and expect a more opinionated deployment with Katello.

ehelms:
I added new enable options rather than carry around old baggage. This then lets things be more selectively enabled as well.

ekohl:
I've been hesitant with adding new options. The katello module determines the enabled Pulp 2 features and the `enable_*` options in this class useless and ignored. Adding these results in 3 options to enable/disable file:
* `--katello-enable-file`
* `--foreman-proxy-content-enable-file` (ignored in the katello scenario)
* `--foreman-proxy-content-enable-pulpcore-file`

There are different ways of exposing this. We can also add it to the scenario as a top level class, resulting in the `--enable-pulpcore-file` option. In the future this will also easily allow us to expose plugin specific options in a discoverable way.

This is what I mentioned when I thought I said I wanted to properly design this. That's why right now I'd reuse the existing options because it'll be less confusing. It's also easier to hide on installs where we don't want to bother the user with Pulp 3 options (too much).

ehelms:
So my problem with the options are they are named `enable_docker` and `enable_iso` which are both not descriptive and not arguably correct. Right now Pulp 3 is entirely controlled by FPC which I think is fine and we should draw the distinction between the two. I also originally kept this from being configurable (and installing file and container) and being more opinionated to allow us to figure the options out later which enabling the use case we know we need to support.

ekohl:
I don't like introducing these options here so in that case I'd prefer to install them unconditionally for now. Then we need to figure out how to properly integrate this later.

ehelms:
Dropped the parameters and installing opinionated set

PR #226, opened 2020-01-06
use pulpcore naming convention

wbclark:
this will require https://github.com/theforeman/puppet-foreman_proxy/pull/556

solves https://github.com/theforeman/puppet-foreman_proxy/issues/550


wbclark:
@ehelms I re-tested this one with https://github.com/theforeman/puppet-foreman_proxy/pull/556 and it looks good. Can you take a look when you have a minute?

PR #228, opened 2020-01-10
fixes #28655 - support fetching files via /pulp/isos with pulp3

wbclark:
This is a WIP. Currently it should disable isos for pulp2 when `$enable_pulpcore_file` but it does not yet add the vhost entry for proxying /pulp/isos to the pulpcore content app

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -40,6 +40,8 @@
 #
 # $pulp_ca_cert::                       Absolute path to PEM encoded CA certificate file, used by Pulp to validate the identity of the broker using SSL.
 #
+# $proxy_pulp_isos_to_pulpcore::        Proxy /pulp/isos to pulpcore at /pulp/content
+#
 # $reverse_proxy::                      Add reverse proxy to the parent
 #
 # $reverse_proxy_port::                 Reverse proxy listening port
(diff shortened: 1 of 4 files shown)


wbclark:
Looking for feedback on the right place to set `$enable_pulpcore_file` to true -- whether that should just be default here or explicitly called in another repo.

wbclark:
Thanks @ehelms . I think the second bullet should be adequately addressed by https://github.com/theforeman/puppet-pulpcore/pull/55 which should also be enabled once `$enable_pulpcore_file` is set to true here.

About the first bullet point, I think it's adequately solved as well if I understand correctly the existing `$enable_iso` and `$enable_file` params. The new `$enable_pulpcore_file` won't affect the existing `$enable_file` but it will override `$enable_iso` to be always false. So IIUC, this leaves the possibility for the pulp2 file module to be enabled but not ISOs specifically. Please let me know if I got that correct.

wbclark:
The latest iteration changes the name of the new parameter to `$proxy_pulp_isos_to_pulpcore` to make it clear that this isn't responsible for whether the file plugin is enabled in pulpcore.

Additionally, the actual ProxyPass logic is moved into this module since the /pulp/isos path is a Katello-ism that doesn't exist in standalone pulp{,core}

wbclark:
Added tests and I believe those should be working as expected after the latest push.

manifests/init.pp
```diff
@@ -12,6 +12,8 @@
 #
 # $enable_file::                        Enable file content plugin
 #
+# $proxy_pulp_isos_to_pulpcore::        Proxy /pulp/isos to pulpcore at /pulp/content 
+#
```
ehelms: I'm thinking this should be in the advanced parameters section as we don't want users to mess with this typically.

wbclark: Makes sense, thanks

PR #229, opened 2020-01-14
Fixes #28654 - support client cert auth with pulp3

wbclark:
Fixes #28654

This works in conjuction with https://github.com/theforeman/puppet-pulpcore/pull/56

### DIFF
--- a/templates/pulpcore-apache.conf.erb
+++ b/templates/pulpcore-apache.conf.erb
@@ -2,3 +2,7 @@
   ProxyPassReverse <%= scope['pulpcore::apache::api_path'] %> <%= scope['pulpcore::apache::api_url'] %>
   ProxyPass <%= scope['pulpcore::apache::content_path'] %> <%= scope['pulpcore::apache::content_url'] %>
   ProxyPassReverse <%= scope['pulpcore::apache::content_path'] %> <%= scope['pulpcore::apache::content_url'] %>
+
+  <Location /pulp/api/v3>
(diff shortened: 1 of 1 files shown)


templates/pulpcore-apache.conf.erb
```diff
@@ -2,3 +2,7 @@
   ProxyPassReverse <%= scope['pulpcore::apache::api_path'] %> <%= scope['pulpcore::apache::api_url'] %>
   ProxyPass <%= scope['pulpcore::apache::content_path'] %> <%= scope['pulpcore::apache::content_url'] %>
   ProxyPassReverse <%= scope['pulpcore::apache::content_path'] %> <%= scope['pulpcore::apache::content_url'] %>
+
+  <Location /pulp/api/v3>
+  RequestHeader set REMOTE_USER "%{SSL_CLIENT_S_DN_CN}s" env=SSL_CLIENT_S_DN
```
ekohl: My impression is that this means that if `SSL_CLIENT_S_DN` is set, `SSL_CLIENT_S_DN_CN` will be passed as the `REMOTE_USER` header. I think that also means that if the user doesn't provide a certificate, they can pass the `REMOTE_USER` header themselves and pretend to be any user. Please verify that we haven't opened a huge security issue here.

jlsherrill: i swear i tried this, but you are correct @ekohl !!  Adding this on the line before seems to resolve it:
```
RequestHeader set REMOTE_USER ""
```

PR #231, opened 2020-01-16
Fixes #28711 - /pulp/content is not being served over http with pulp3

wbclark:
Adds http support for /pulp/content using foreman::config::apache::fragment


templates/pulpcore-apache-http.conf.erb
```diff
@@ -0,0 +1,2 @@
+  ProxyPass <%= scope['pulpcore::apache::content_path'] %> <%= scope['pulpcore::apache::content_url'] %>
+  ProxyPassReverse <%= scope['pulpcore::apache::content_path'] %> <%= scope['pulpcore::apache::content_url'] %>
```
ehelms: Is the indent intended?

ehelms:
I think looking at this it'd be worth splitting https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/templates/pulpcore-apache.conf.erb into two pieces. One fragment for the Content app and one for the API app. Then you can create fragments for each and re-use the definitions of the Content app for http and https.

wbclark:
> I think looking at this it'd be worth splitting https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/templates/pulpcore-apache.conf.erb into two pieces. One fragment for the Content app and one for the API app. Then you can create fragments for each and re-use the definitions of the Content app for http and https.

This approach seems much cleaner. Will test it.

spec/classes/foreman_proxy_content_spec.rb
```diff
@@ -60,8 +60,10 @@ class { 'foreman_proxy::plugin::pulp':
         it { is_expected.to contain_class('pulpcore').with(manage_apache: false) }
 
         it do
-          is_expected.to contain_foreman__config__apache__fragment('pulpcore')
+          is_expected.to contain_foreman__config__apache__fragment('pulpcore-https')
             .with_ssl_content(%r{ProxyPass /pulp/api/v3 http://127\.0\.0\.1:24817/pulp/api/v3})
+          is_expected.to contain_foreman__config__apache__fragment('pulpcore-http')
+            .with_content(%r{ProxyPass /pulp/content http://127\.0\.0\.1:24816/pulp/content})
```
ehelms: Can you add this `with_content` check two lines above as well? This way the tests ensure that the content app proxy is in http and https fragments.

wbclark: @ehelms Done. I also added a test for the `pulpcore-isos` fragment and renamed its template to be consistent with the naming for content and api templates. Thanks for hte suggestion.

ekohl:
Nit on the commit message: she short summary should say what it does, not what it doesn't do. https://chris.beams.io/posts/git-commit/#imperative has a great description and also examples.

wbclark:
> Nit on the commit message: she short summary should say what it does, not what it doesn't do. https://chris.beams.io/posts/git-commit/#imperative has a great description and also examples.

Noted. I simply copied the title of the Redmine but since it's linked by the issue # already I'll instead reword it to be more useful.

PR #232, opened 2020-01-20
katello/certs 7.x compatibility & use certs::qpid_client variables

ekohl:
certs 7.0.0 started to create qpid files owned by the pulp group. This group is created by pulp-server which is in pulp::install. This more specific chaining avoids dependency cycles.

qpid_client_cert is the same:
   $qpid_client_cert      = $certs::qpid_client_cert,

The CA is a different path:
   $qpid_client_ca_cert   = $certs::qpid_client_ca_cert,

Which is defined as:
   $qpid_client_ca_cert = "${pulp_pki_dir}/qpid/ca.crt"

This is different than what it used to be:
   $ca_cert = "${pki_dir}/certs/${default_ca_name}.crt"

The contents should match.


ekohl:
This is an alternative to theforeman/puppet-certs#270. Related to theforeman/puppet-certs#272.

PR #233, opened 2020-01-20
Fixes #28696 - configures apache to serve docker content with pulp3

wbclark:


templates/pulpcore-docker-apache.conf.erb
```diff
@@ -0,0 +1,6 @@
+<Location /pulpcore_registry/v2/>
+  SSLRequire %{SSL_CLIENT_S_DN_CN} eq "admin" 
+</Location>
+
+ProxyPass /pulpcore_registry/v2/ http://127.0.0.1:24816/v2/
```
wbclark: I'm not sure if the trailing `/` here is necessary. I included it because it is written that way on Redmine #28696 but the same was absent from `ProxyPassReverse` and similar `ProxyPass` declarations for other paths.

ekohl: This should take the host and port into account as the other templates do

> I'm not sure if the trailing `/` here is necessary. I included it because it is written that way on Redmine #28696 but the same was absent from `ProxyPassReverse` and similar `ProxyPass` declarations for other paths.

Probably a good thing. In theory you can get odd things with `/v21` otherwise. I do think `ProxyPass` and `ProxyPassReverse` should match though.

wbclark: Point taken. I don't see a need to introduce new parameters for the registry path in puppet-pulpcore so the template here will look a little different from the others but it now depends on the `content_host` and `content_port`.

wbclark: Point taken. I don't see a need to introduce new parameters for the registry path in puppet-pulpcore so the template here will look a little different from the others but it now depends on the `content_host` and `content_port`. Thanks!

PR #235, opened 2020-01-21
Do not configure Pulp CA certificate

ehelms:
We do not configure this on the main server since this CA is used for consumer actions in Pulp 2 which we do not use that functionlity.


ehelms:
This should fix nightly install pipeline.

ekohl:
Looks like 54daeac19cf59da34b7cab6d6e54770b68b8dd2d introduced this. Looking at puppet-pulp then we should actually pass in `https_ca_cert` now. If not passed in, it uses `ca_cert` which will now no longer match the actual cert and probably breaks client cert authentication. I think https://github.com/theforeman/puppet-foreman_proxy_content/pull/236 should fix this so let's continue the discussion there.

PR #237, opened 2020-01-30
Refs #28904 - configure pulpcore django_remote_user

synkd:


manifests/init.pp
```diff
@@ -293,8 +293,9 @@
     include foreman
 
     class { 'pulpcore':
-      manage_apache => false,
-      servername    => $foreman::servername,
+      django_remote_user => 'HTTP_REMOTE_USER',
```
ehelms: Parameter name needs update now

PR #238, opened 2020-02-05
Refs #28720 - connect to mongo for content migrations

wbclark:
Looking for feedback on the approach of conditionally including `katello::pulp` and forming an array from the mongo db parameters. If that looks good I'll add tests. Thanks!

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -298,6 +298,22 @@ class foreman_proxy_content (
       servername               => $foreman::servername,
     }
 
+    if $pulp_master {
+      include pulp
+      class { 'pulpcore::plugin::migration':
+        mongo_db_name         => $pulp::db_name,
+        mongo_db_seeds        => $pulp::db_seeds,
+        mongo_db_username     => $pulp::db_username,
+        mongo_db_password     => $pulp::db_password,
+        mongo_db_replica_set  => $pulp::db_replica_set,
+        mongo_db_ssl          => $pulp::db_ssl,
+        mongo_db_ssl_keyfile  => $pulp::db_ssl_keyfile,
+        mongo_db_ssl_certfile => $pulp::db_ssl_certfile,
+        mongo_db_verify_ssl   => $pulp::db_verify_ssl,
+        mongo_db_ca_path      => $pulp::db_ca_path,
+      }
+    }
+
     include pulpcore::plugin::container
     include pulpcore::plugin::file
 
--- a/spec/classes/foreman_proxy_content_spec.rb
+++ b/spec/classes/foreman_proxy_content_spec.rb
@@ -71,6 +71,38 @@ describe 'foreman_proxy_content' do
         end
       end
 
+      context 'pulp-2to3-migration' do
+        let(:params) do
+          {
+            qpid_router: false
+          }
+        end
+
+        let(:pre_condition) do
+          <<-PUPPET
+          include foreman_proxy
+          class { 'foreman_proxy::plugin::pulp':
+            enabled          => true,
+            pulpnode_enabled => false,
+            pulpcore_enabled => true,
+            pulpcore_mirror  => false,
+          }
+          PUPPET
+        end
+
+        it { is_expected.to compile.with_all_deps }
+        it { is_expected.to contain_class('pulpcore').with(manage_apache: false) }
+
+        it do
+          is_expected.to contain_class('pulpcore::plugin::migration')
+            .with_mongo_db_name('pulp_database')
+            .with_mongo_db_seeds('localhost:27017')
+            .with_mongo_db_ssl(false)
+            .with_mongo_db_verify_ssl(true)
+            .with_mongo_db_ca_path('/etc/pki/tls/certs/ca-bundle.crt')
+        end
+      end
+
       context 'with rhsm_hostname and rhsm_url' do
         let(:params) do
           {


manifests/init.pp
```diff
@@ -292,10 +292,30 @@
...
+      }
+    } else {
+      $mongo_db_config = {}
+    }
+
     class { 'pulpcore':
       remote_user_environ_name => 'HTTP_REMOTE_USER',
       manage_apache            => false,
       servername               => $foreman::servername,
+      *                        => $mongo_db_config,
```
ehelms: I do see the value that @ekohl mentioned with this set of parameters living on the `migration` class and declaring that directly rather than as an includes. It encapsulates the migration portion a bit cleaner. What do you think @wbclark ?

wbclark: I agree, there is no reason to have the parameters outside of the `migration` plugin. I've opened https://github.com/theforeman/puppet-pulpcore/pull/63 to move them and will push a new version of this commit which respects that change.

wbclark: @ehelms I've updated this for https://github.com/theforeman/puppet-pulpcore/pull/63

manifests/init.pp
```diff
@@ -292,6 +292,26 @@
...
+        'mongo_db_replica_set'  => $katello::pulp::mongodb_replica_set,
+        'mongo_db_ssl'          => $katello::pulp::mongodb_ssl,
+        'mongo_db_ssl_keyfile'  => $katello::pulp::mongodb_ssl_keyfile,
+        'mongo_db_ssl_certfile' => $katello::pulp::mongodb_ssl_certfile,
+        'mongo_db_verify_ssl'   => $katello::pulp::mongodb_verify_ssl,
+        'mongo_db_ca_path'      => $katello::pulp::mongodb_ca_path,
+      }
+
+      class { 'pulpcore::plugin::migration':
+        * => $mongo_db_config,
```
ehelms: Is there an advantage to the `*` vs. just setting each parameter on the class itself?

wbclark: Nope, it's just an artifact of the previous approach I was using. It's no longer necessary, so I removed it.

ehelms:
Thinking more on it, a parameter that enabled the migration plugin and its connection to MongoDB woul d be useful. On a fresh install, Pulp 3 does not need to connect to MongoDB.

wbclark:
> Thinking more on it, a parameter that enabled the migration plugin and its connection to MongoDB woul d be useful. On a fresh install, Pulp 3 does not need to connect to MongoDB.

My thinking is that we have it conditional on `$pulp` already, which should eventually be `false` for fresh installations once pulpcore only is default -- exactly the circumstances where the migration plugin is not needed. Will that be sufficient?

ehelms:
> Thanks for the feedback @ehelms . My thinking is that we have it conditional on `$pulp` already, which should eventually be `false` for fresh installations once pulpcore only is default -- exactly the circumstance where the migration plugin is not needed. Will that be sufficient?

After talking a bit to Pulp 3 integration team, we can leave it as you have it and establish the connection from the outset.

manifests/init.pp
```diff
@@ -292,6 +292,22 @@
   if $pulpcore and !$pulpcore_mirror {
     include foreman
 
+    if $pulp {
```
ekohl: It shouldn't matter, but I would move this below the pulpcore definition. That way all plugins are defined after the class itself and keeps them a bit grouped. In the future the plugin define might include pulpcore and then the current implementation will fail to compile due to a duplicate class definition.

spec/classes/foreman_proxy_content_spec.rb
```diff
@@ -71,6 +71,40 @@ class { 'foreman_proxy::plugin::pulp':
...
+        let(:params) do
+          {
+            qpid_router: false
+          }
+        end
+
+        let(:pre_condition) do
+          <<-PUPPET
+          include foreman_proxy
+          include katello::pulp
```
ekohl: This shouldn't be needed since the code also has an include
```suggestion
```

manifests/init.pp
wbclark: @ekohl done.

spec/classes/foreman_proxy_content_spec.rb
wbclark: @ekohl in fact I had to remove that include since it was causing conflicts when both pulp2 and pulpcore are enabled (we are already declaring pulp). I'm now getting the mongo params from the already declared pulp class. I'll remove the dependencies for puppet-katello now since they are no longer needed.

wbclark:
> > Thanks for the feedback @ehelms . My thinking is that we have it conditional on `$pulp` already, which should eventually be `false` for fresh installations once pulpcore only is default -- exactly the circumstance where the migration plugin is not needed. Will that be sufficient?
> 
> After talking a bit to Pulp 3 integration team, we can leave it as you have it and establish the connection from the outset.

OK, great. Thanks for the follow up @ehelms. If it later comes up that we need to decouple enabling the plugin from the `pulp` and `pulpcore` parameters and introduce a new param instead, we can revisit.

manifests/init.pp
```diff
@@ -298,6 +298,21 @@
       servername               => $foreman::servername,
     }
 
+    if $pulp {
```
ehelms: This should actually be `if $pulp_master` as `$pulp` denotes an external proxy with content. And note that with a master Pulp is configured by Katello so its not clear to me if the parameters below would be available. Perhaps include on pulp would make them so?

spec/classes/foreman_proxy_content_spec.rb
```diff
@@ -71,6 +71,39 @@ class { 'foreman_proxy::plugin::pulp':
...
+            pulpnode_enabled => false,
+            pulpcore_enabled => true,
+            pulpcore_mirror  => false,
+          }
+          PUPPET
+        end
+
+        it { is_expected.to compile.with_all_deps }
+        it { is_expected.to contain_class('pulp').with(manage_squid: true) }
+        it { is_expected.to contain_class('pulpcore').with(manage_apache: false) }
```
ehelms: I'm not clear how this test and the test above it are related to the migration code.

wbclark: @ehelms they aren't. My thinking is that we might as well test that pulp and pulpcore continue to work in the case where both are present and we are turning the migration plugin on. In fact since puppet-katello handles pulp setup for the main katello server, I should definitely remove the pulp test here. I can remove the pulpcore test as well if you feel it is redundant.

PR #240, opened 2020-02-11
Fixes #28983 - Allow integers for $proxy_pass_params

ekohl:
f95a6eef1d510e65e2499367d738935b151cd05c added a parameter to allow setting parameters to proxy_pass but only accepted strings. When setting this via hiera, values like TTL will usually end up as integers and are valid. This can be worked around by quoting them in hiera, but it's a bad user experience.


ehelms:
Could we get a Redmine issue?

ekohl:
Updated with an issue. Also mentioned it can be worked around by explicitly quoting the values to make them strings.

PR #241, opened 2020-02-11
Refs #28901 - support external postgres database in pulpcore

wbclark:
This depends on https://github.com/theforeman/puppet-pulpcore/pull/67

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -88,6 +88,16 @@
 #                                       mongo database has slow I/O, then setting a higher number may resolve issues where workers are
 #                                       going missing incorrectly.
 #
+# $pulpcore_manage_postgresql::           Manage the pulpcore postgresql database
+#
+# $pulpcore_postgresql_host::             Host of the pulpcore postgresql database. Must be specified if external/unmanaged.
+#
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -296,6 +296,11 @@
       remote_user_environ_name => 'HTTP_REMOTE_USER',
       manage_apache            => false,
       servername               => $foreman::servername,
+      postgresql_manage_db     => $foreman::db_manage,
```
ekohl: IMHO these should not be the same ones as Foreman. They're independent. We have separate parameters for candlepin as well.

manifests/init.pp
```diff
@@ -88,6 +88,16 @@
 #                                       mongo database has slow I/O, then setting a higher number may resolve issues where workers are
 #                                       going missing incorrectly.
 #
+# $pulpcore_manage_postgres::           Manage the pulpcore postgresql database
```
ehelms: I am thinking we might as should just expand this out to `postgresql` for references to use the full name

wbclark: @ehelms done

PR #243, opened 2020-02-14
Use loose coupling to Foreman's Apache config

ekohl:
By only including the Apache config, we get only what we really need. This depends on https://github.com/theforeman/puppet-foreman/pull/800

Also uses a single Apache fragment for both HTTP and HTTPS.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -338,16 +338,13 @@ class foreman_proxy_content (
     include pulpcore::plugin::container
     include pulpcore::plugin::file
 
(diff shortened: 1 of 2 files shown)


ehelms:
Does https://github.com/theforeman/puppet-katello/blob/master/manifests/pulp.pp#L99 need the same treatment?

ekohl:
Yes, I think that's another candidate.

ehelms:
@ekohl I introduced a rebase needed here

ekohl:
:green_apple:

PR #244, opened 2020-02-20
Fixes #28901 - Support SSL connection for external Pulpcore PostgreSQL…

wbclark:
… database

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -52,9 +52,14 @@ class foreman_proxy_content::params {
 
   $manage_broker                = true
 
-  $pulpcore_manage_postgresql   = true
-  $pulpcore_postgresql_host     = 'localhost'
-  $pulpcore_postgresql_port     = 5432
-  $pulpcore_postgresql_user     = 'pulp'
-  $pulpcore_postgresql_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32))
+  $pulpcore_manage_postgresql      = true
+  $pulpcore_postgresql_host        = 'localhost'
+  $pulpcore_postgresql_port        = 5432
+  $pulpcore_postgresql_user        = 'pulp'
+  $pulpcore_postgresql_password    = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32))
+  $pulpcore_postgresql_ssl         = false
+  $pulpcore_postgresql_ssl_require = true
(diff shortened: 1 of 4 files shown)


ekohl:
Depends on https://github.com/theforeman/puppet-pulpcore/pull/71. If we want to do it right, we should update the version in puppet-pulpcore's metadata.json to increase the minor and in this metadata.json increase the minimum version of puppet-pulpcore to that.

wbclark:
> Depends on [theforeman/puppet-pulpcore#71](https://github.com/theforeman/puppet-pulpcore/pull/71). If we want to do it right, we should update the version in puppet-pulpcore's metadata.json to increase the minor and in this metadata.json increase the minimum version of puppet-pulpcore to that.

Thanks. I have made both changes.

manifests/params.pp
```diff
@@ -2,59 +2,64 @@
...
-  $enable_deb                   = true
-
-  $manage_broker                = true
-
-  $pulpcore_manage_postgresql   = true
-  $pulpcore_postgresql_host     = 'localhost'
-  $pulpcore_postgresql_port     = 5432
-  $pulpcore_postgresql_user     = 'pulp'
-  $pulpcore_postgresql_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32))
+  $parent_fqdn                     = $facts['fqdn']
```
ekohl: I've started to stop aligning all parameters because it creates a lot of noise in git logs. Feel free to do the same.

wbclark: @ekohl I do find it's very nice to have them aligned for readability. Is it a better approach to do these alignment fixups in a separate commit in the future?

wbclark: Another thought, is it better still to keep the consistent alignment only within each block, when the number of parameters is large?

ekohl: yes, I think per block alignment makes much more sense and is a decent compromise between the two

metadata.json
```diff
@@ -35,6 +35,10 @@
     {
       "name": "katello/certs",
       "version_requirement": ">= 6.0.0 < 8.0.0"
+    },
+    {
+      "name": "theforeman/pulpcore",
```
ekohl: I noticed this too after the release and we should have done this before already.

wbclark:
> Code wise +1 but I think this commit should be `Fixes` instead of `Refs`

Updated commit message and PR title. @ehelms do we need to get pulpcore 0.2.0 out to puppet forge before this can be merged?

ekohl:
No, nightly doesn't need that since it pulls in the git version.

PR #245, opened 2020-03-06
Fixes #29278: Use default CA for crane SSL chain

ehelms:


manifests/init.pp
```diff
@@ -252,6 +252,7 @@
       cert         => $certs::apache::apache_cert,
       key          => $certs::apache::apache_key,
       ca_cert      => $certs::katello_server_ca_cert,
+      ssl_chain    => $certs::katello_default_ca_cert,
```
jlsherrill: i think the reverse is done for the 443 port?

jlsherrill: Yeah, chain is set to server, ca is set to default.  I'm not saying thats correct, just that they don't match.
  SSLCertificateChainFile "/etc/pki/katello/certs/katello-server-ca.crt"
  SSLCACertificateFile    "/etc/pki/katello/certs/katello-default-ca.crt"

ehelms: Ahh yea, fair. Generally we are handling this wrong I think and looking ahead to things like EL8 which has newer versions of Apache we should drop SSLCertificateChainFile as it is deprecated in Apache 2.4.8.

I have swapped these.

ekohl: The thing is, there is no replacement for the behavior we need: a CA to serve for the certificate and a different CA for the client authentication.

PR #246, opened 2020-03-09
Fixes #29278: Use default CA for crane SSL chain

ekohl:
(cherry picked from commit 85f85fc1b12c693101156559708c5a855113ff47)

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -11,14 +11,20 @@ fixtures:
       repo: "https://github.com/puppetlabs/puppetlabs-cron_core"
       puppet_version: ">= 6.0.0"
     datacat:       "https://github.com/richardc/puppet-datacat"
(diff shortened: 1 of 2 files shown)


evgeni:
`Resource type not found: Redis::RedisUrl` :(

evgeni:
(ACK on the change, tho, just make Travis happy please :) )

ehelms:
@ekohl could you tackle this for me please

philipp1992:
hi,

we are facing the same issue with custom ssl certs and katello 3.15 + foreman 2.0.
Ive tried to apply that patch to the manifest/init.pp file, but when i re-run the foreman-installer, i get an error saying, that there is no cache file for init.pp and no puppet module parser.

Any hints on how to apply this fix to a foreman 2.0 installation without upgrading to the nightly build?

PR #251, opened 2020-04-23
Refs #29214: Only install Pulp and Qpid on El7

ehelms:


ehelms:
Is there a clean way to include parameters based on the OS version only? While failing is nice, in the installer use case I wouldn't want any of these showing up to the users.

ekohl:
No, but those options have been available for a long time and users don't appear to be using them that much anyway. I'd propose we fix make sure it works correctly, then look at trimming down options. We'll drop Pulp 2 soon anway.

manifests/init.pp
```diff
@@ -214,7 +218,7 @@
     }
   }
 
-  if $pulp_master or $pulp {
+  if ($pulp_master or $pulp) and $facts['os']['release']['major'] == '7' {
```
ekohl, line 221: All these checks are now redundant since we fail early on.

ehelms: Safety first!

ehelms: That is me saying, the fail does cover it but I also see this has helping to cover our bases and make it clear in the code this block is EL7 only. I debated moving a bunch of htis to their own class to highlight that as well but didn't want to do too much re-factoring.

ekohl: Feels a bit redundant to me since you know there's no Pulp 2 on EL8 anyway.

ehelms: I know that, but doesn't mean everyone reading this code will clearly know that.

ekohl: Partly I agree, but the code is already very hard to follow. Adding more conditionals just makes it harder rather than easier IMHO. I'm waiting till we can drop Pulp for a big refactor.

PR #252, opened 2020-04-28
Fixes #29589 - Ensure pulpcore before proxy

ekohl:
When the Foreman Proxy registration happens, Pulp's features are queried and presented as capabilities. If the service is down, Katello can't detect the right content types.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -338,6 +338,7 @@ class foreman_proxy_content (
       postgresql_db_ssl_cert    => $pulpcore_postgresql_ssl_cert,
       postgresql_db_ssl_key     => $pulpcore_postgresql_ssl_key,
       postgresql_db_ssl_root_ca => $pulpcore_postgresql_ssl_root_ca,
+      before                    => Class['foreman_proxy::plugin::pulp'],
     }
 
     if $pulp_master {
--- a/spec/classes/foreman_proxy_content_spec.rb
+++ b/spec/classes/foreman_proxy_content_spec.rb
@@ -57,7 +57,7 @@ describe 'foreman_proxy_content' do
         end
 
         it { is_expected.to compile.with_all_deps }
-        it { is_expected.to contain_class('pulpcore').with(manage_apache: false) }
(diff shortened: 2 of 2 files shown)


ehelms:
Looks sane but that dependency cycle is funky with the daemon reloads chaining together.

ekohl:
Yes. That isn't needed in Puppet >= 6.1 but for now we have to support Puppet 5 too. The workaround is to not let `systemd::unit_file` manage the service.

manifests/init.pp
```diff
@@ -338,6 +338,7 @@
       postgresql_db_ssl_cert    => $pulpcore_postgresql_ssl_cert,
       postgresql_db_ssl_key     => $pulpcore_postgresql_ssl_key,
       postgresql_db_ssl_root_ca => $pulpcore_postgresql_ssl_root_ca,
+      before                    => Class['foreman_proxy::plugin::pulp'],
```
ehelms, line 341: Could we do it before the register?

```
iff --git a/manifests/init.pp b/manifests/init.pp
index 9bea8db..78d833a 100644
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -338,6 +338,7 @@ class foreman_proxy_content (
       postgresql_db_ssl_cert    => $pulpcore_postgresql_ssl_cert,
       postgresql_db_ssl_key     => $pulpcore_postgresql_ssl_key,
       postgresql_db_ssl_root_ca => $pulpcore_postgresql_ssl_root_ca,
+      before                    => Class['foreman_proxy::register'],
     }
 
     if $pulp_master {
```

ekohl: Tried that, same error.

ehelms: Strange, I made this change, ran tests. Everything passed. I ran it again, got these failures. Perhaps my fixtures weren't up to date? Implies it was a more recent change that led to this?

ekohl: Yes, https://github.com/theforeman/puppet-puppet/commit/4302973e8f47c77ed35b2a399c1ccef2d503b400

ekohl:
https://github.com/theforeman/puppet-puppet/pull/741 would remove the dependency cycle. Once that's merged, tests can be restarted.

PR #255, opened 2020-05-06
Fixes #29214: Add EL8 support to metadata

ehelms:


mmoll:
Travis says no. ;)

ehelms:
Travis says yesss

spec/classes/foreman_proxy_content__broker_spec.rb
```diff
@@ -7,7 +7,9 @@
         facts
       end
 
-      it { is_expected.to compile.with_all_deps }
```
ekohl, line 10: Given this example, normally I'd be cleaner to do modify line 5:
```ruby
context "on #{os}", if: facts[:operatingsystemmajrelease] == '7'
```

That skips the entire context and produces less output.

spec/classes/foreman_proxy_content__pub_dir_spec.rb
```diff
@@ -7,6 +7,16 @@
         facts
       end
 
+      let(:pre_condition) do
```
ekohl, line 10: Looking at this and the actual code, I wonder why this is here:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/8f76e215534db857228b56b075598d39afcd0fbe/manifests/pub_dir.pp#L14

If you remove that, the class becomes standalone and this isn't needed.

@timogoebel you added this in c363e7c0b068e7d2d06c5c0698c98f545c1c1f31 but I don't see why.

spec/classes/foreman_proxy_content_spec.rb
```diff
@@ -6,36 +6,48 @@
...
-        it { is_expected.not_to contain_class('pulpcore') }
-        it { is_expected.to contain_class('foreman_proxy_content::pub_dir') }
+        it { is_expected.to contain_package('katello-debug') }
+      end
 
-        it do
-          is_expected.to contain_pulp__apache__fragment('gpg_key_proxy')
-            .with_ssl_content(%r{ProxyPass /katello/api/v2/repositories/ https://foo\.example\.com/katello/api/v2/repositories/})
+      if facts[:operatingsystemmajrelease] == '7'
+        context 'with pulp' do
```
ekohl, line 24: Like in the first comment, you can write less indented code by using `if:` here
```suggestion
      context 'with pulp', if: facts[:operatingsystemmajrelease] == '7' do
```

spec/classes/foreman_proxy_content__pub_dir_spec.rb
timogoebel: I don't remember. Probably because the class needs some prerequisites the other class sets up to work. Maybe files referenced in the `additional_includes` section?

ehelms: Could we opt to file an issue and come back to it when the Pulp cleanup happens? I'm wanting to be minimally invasive

spec/classes/foreman_proxy_content_spec.rb
```diff
@@ -141,47 +157,60 @@ class { 'foreman_proxy::plugin::pulp':
         end
       end
 
+      if facts[:operatingsystemmajrelease] == '7'
+        context 'pulp-2to3-migration' do
```
ekohl, line 161: I'd also prefer if this used `if:` to reduce the indenting.

ehelms:
We do need to support migration plugin on EL8 with pulpcore. Are you seeing a spot where I prohibited this in the tests?

wbclark:
> We do need to support migration plugin on EL8 with pulpcore. Are you seeing a spot where I prohibited this in the tests?

Not at all, my question is whether there should be a new separate test (EL8 only) for pulpcore and migration plugin, without pulp.

ehelms:
Ah ha! Fair point, I will add that.

ehelms:
I looked into thi smore and I realized that right now the code is tied to the pulp class itself rather than inputs to foreman_proxy_content and will require some re-factor/design to support migration on EL8 talking to a Pulp on El7. I would like to punt that to a later feature update and leave this PR as is.

wbclark:
> I looked into thi smore and I realized that right now the code is tied to the pulp class itself rather than inputs to foreman_proxy_content and will require some re-factor/design to support migration on EL8 talking to a Pulp on El7. I would like to punt that to a later feature update and leave this PR as is.
> 
> Edit: https://projects.theforeman.org/issues/29723

Great, agreed then

PR #257, opened 2020-05-15
Allow EL8 supported versions of dependencies

wbclark:


mmoll:
merged, thanks @wbclark!

PR #261, opened 2020-06-04
Refs #30023: Enable Pulpcore RPM plugin

ehelms:


manifests/init.pp
```diff
@@ -149,6 +151,7 @@
   Boolean $enable_yum = $foreman_proxy_content::params::enable_yum,
   Boolean $enable_file = $foreman_proxy_content::params::enable_file,
   Boolean $proxy_pulp_isos_to_pulpcore = $foreman_proxy_content::params::proxy_pulp_isos_to_pulpcore,
+  Boolean $proxy_pulp_yum_to_pulpcore = $foreman_proxy_content::params::proxy_pulp_yum_to_pulpcore,
```
wbclark, line 154: The value here is defined to reference the definition in params.pp, but the definition is absent from params.pp

manifests/init.pp
```diff
@@ -363,6 +366,7 @@
 
     include pulpcore::plugin::container
     include pulpcore::plugin::file
+    include pulpcore::plugin::rpm
```
evgeni, line 369: general question: shouldn't those plugin includes (at least file and rpm) be inside the respective "proxy to pulpcore" blocks? as otherwise we'd be installing the pulpcore plugin but not using it, which seems like a waste?

ekohl: The idea is that the plugin is installed. This allows you to migrate content from Pulp 2. Also the Pulp 3 content URLs are always available. The "proxy to pulpcore" blocks are about routing the old Pulp 2 URLs to Pulp 3 so clients don't have to be modified. Ideally they should, but this allows to do so at a convenient pace.

manifests/init.pp
```diff
@@ -287,7 +290,7 @@
 
     class { 'pulp':
       enable_ostree          => $enable_ostree,
-      enable_rpm             => $enable_yum,
+      enable_rpm             => $enable_yum and !$proxy_pulp_yum_to_pulpcore,
```
ekohl, line 293: Not really for this PR, but this made me think of it: we don't have conditional to disable crane if we proxy container content to Pulpcore.

ehelms: Yea, I think we just were leaning towards leave Crane alone till we can safely rip it out.

PR #263, opened 2020-06-08
Fixes #30059 - Add config for protected content

sjha4:


jlsherrill:
we also need to make sure the plugin gets installed, so we probably need a new class here:  https://github.com/theforeman/puppet-pulpcore/tree/master/manifests/plugin  as well as a reference to it in this module.

sjha4:
https://github.com/theforeman/puppet-pulpcore/pull/99 is merged.

sjha4:
@wbclark : We would always want to install certguard on both katello server and content proxies where pulpcore is enabled.

sjha4:
On the server, the plugin would be automatically picked up and installed with the file [here](https://github.com/theforeman/puppet-pulpcore/blob/master/manifests/plugin/certguard.pp ): No?

wbclark:
> On the server, the plugin would be automatically picked up and installed with the file [here](https://github.com/theforeman/puppet-pulpcore/blob/master/manifests/plugin/certguard.pp): No?

This defines a class `pulpcore::plugin::certguard` but this class is not (yet) called.

The proper place to do that would actually be to include it in this module right before https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/init.pp#L367-L369

templates/pulpcore-content-apache.conf.erb
```diff
@@ -1,2 +1,8 @@
-  ProxyPass <%= scope['pulpcore::apache::content_path'] %> <%= scope['pulpcore::apache::content_url'] %>
-  ProxyPassReverse <%= scope['pulpcore::apache::content_path'] %> <%= scope['pulpcore::apache::content_url'] %>
+ProxyPass <%= scope['pulpcore::apache::content_path'] %> <%= scope['pulpcore::apache::content_url'] %>
+ProxyPassReverse <%= scope['pulpcore::apache::content_path'] %> <%= scope['pulpcore::apache::content_url'] %>
+
+<Location /pulp/content>
+RequestHeader unset X-CLIENT-CERT
+RequestHeader set X-CLIENT-CERT "%{SSL_CLIENT_CERT}s"
+</Location>
+
```
wbclark, line 8: Is it intentional/beneficial to add a new blank line to the end of these 3 templates?

sjha4: Updated..newline isn't needed here.

PR #264, opened 2020-06-11
Refs #30019 - Move Pulpcore PostgreSQL params to basic params

wbclark:
This should cause the parameters to be displayed by `foreman-installer --help` without requiring `--foreman-installer --full-help`

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -18,6 +18,26 @@
 #
 # $enable_deb::                         Enable debian content plugin
 #
+# $pulpcore_manage_postgresql::         Manage the Pulpcore PostgreSQL database.
+#
+# $pulpcore_postgresql_host::           Host of the Pulpcore PostgreSQL database. Must be specified if external/unmanaged.
+#
+# $pulpcore_postgresql_port::           Port of the Pulpcore PostgreSQL database.
+#
+# $pulpcore_postgresql_user::           User of the Pulpcore PostgreSQL database.
+#
+# $pulpcore_postgresql_password::       Password of the Pulpcore PostgreSQL database.
+#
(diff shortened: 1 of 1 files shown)


ehelms:
I think I disagree with this change as these are what I would consider `advanced` parameters. But I would like to see what others think. @ekohl @evgeni

ehelms:
Adding a note that Candlepin DB params are advanced currently: https://github.com/theforeman/puppet-katello/blob/master/manifests/init.pp#L37

I think what we really need is a parameter to the installer that allows configuring these options and setting it for all the services given Foreman/Candlepin/Pulp 3 all use a single database instance now.

wbclark:
> I think what we really need is a parameter to the installer that allows configuring these options and setting it for all the services given Foreman/Candlepin/Pulp 3 all use a single database instance now.

Now that Foreman, Pulp, and Candlepin are all competing for PostgreSQL connections, I could see it being desirable for large deployments to use multiple DB instances, so that heavy registration traffic is not bottlenecking pulp performance on DB transactions.

wbclark:
To follow up, I've started a discussion on that topic at https://community.theforeman.org/t/split-external-dbs/19336

ekohl:
I'd suggest a `Pulpcore PostgreSQL` or `Pulpcore Remote Database` group in the docs. Due to https://projects.theforeman.org/issues/29788 it still won't show up in `--help`, but it'd be a step in the right direction. IMHO remote DB is still advanced.

Also note that you can do `man foreman-installer` but it shows the options of the `foreman` scenario. I've been thinking about introducing `foreman-installer-scenario-$scenario` and making `man foreman-installer` only a generic / basic man page. That way you don't have to wait for `--help` / `--full-help` if you just want to know which options are there and don't care about the values.

wbclark:
Closing this while we are having an ongoing discussion about how to best present installer params to the user.

PR #266, opened 2020-06-25
Fixes #30217 - configure pulpcore db name

wbclark:
I realized when installing Katello with external DBs that we have the option to do this for Foreman and Candlepin DBs but not for Pulpcore


manifests/params.pp
```diff
@@ -58,6 +58,7 @@
   $pulpcore_postgresql_port        = 5432
   $pulpcore_postgresql_user        = 'pulp'
   $pulpcore_postgresql_password    = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32))
+  $pulpcore_postgresql_db_name     = 'Pulpcore'
```
ekohl, line 61:
```suggestion
  $pulpcore_postgresql_db_name     = 'pulpcore'
```

PR #270, opened 2020-07-02
Fixes #30057 - Expose Pulpcore Worker Count in Puppet-FPC

wbclark:


wbclark:
> This was added in pulpcore 1.1 and metadata.json should reflect this.

I don't think it was in before 1.1 release. How to handle it in that case?

ekohl:
You require `>= 1.1.0` in metadata.json.

wbclark:
> You require `>= 1.1.0` in metadata.json.

OK, done

PR #272, opened 2020-07-02
Fixes #30316: Move bootstrap RPM generation from puppet-certs

ehelms:


ehelms:
After this goes in, https://github.com/theforeman/puppet-certs/pull/290 can be merged to clean up but this changed does not require that change.

manifests/bootstrap_rpm.pp
```diff
@@ -0,0 +1,61 @@
...
+  $rhsm_port                         = 443,
+  $rhsm_ca_dir                       = '/etc/rhsm/ca',
+  $candlepin_cert_rpm_alias_filename = 'katello-ca-consumer-latest.noarch.rpm',
+  $katello_www_pub_dir               = '/var/www/html/pub',
+  $katello_server_ca_cert            = $certs::katello_server_ca_cert,
+  $server_ca_name                    = $certs::server_ca_name,
+  $ca_cert                           = $certs::ca_cert,
+  $server_ca                         = $certs::server_ca,
+) {
+  include certs
```
ekohl, line 16: This doesn't really work. The class parameters are compiled and determined before the body so at this point `$katello_server_ca_cert` may be empty.

manifests/bootstrap_rpm.pp
```diff
@@ -0,0 +1,61 @@
+# This class builds an RPM containing the bootstrap for a subscription-manager consumer
+# This file is placed in $katello_www_pub_dir.
+# @api private
+class foreman_proxy_content::bootstrap_rpm (
+  $hostname                          = $facts['networking']['fqdn'],
```
ekohl, line 5: Can we use data types for the parameters to figure out what they are?

Also, I don't think we need to align parameters. I've stopped doing that because it quickly becomes too wide if you also add data types.

manifests/init.pp
```diff
@@ -205,11 +205,10 @@
     notify   => Service['foreman-proxy'],
   }
 
-  class { 'certs::katello':
+  class { 'foreman_proxy_content::bootstrap_rpm':
     hostname       => $rhsm_hostname,
     deployment_url => $rhsm_url,
```
ekohl, line 210: As suggested elsewhere, we can drop the parameter and keep it as a private class parameter. It can be adjusted via hiera if needed, but then it's not part of the public API.

manifests/rhsm_reconfigure_script.pp
```diff
@@ -0,0 +1,57 @@
+define foreman_proxy_content::rhsm_reconfigure_script($ca_cert, $server_ca_cert) {
```
ekohl, line 1: I'm wondering if this needs to be a define. We only use it once AFAIK so might as well be a class, right?

```suggestion
# @summary Generate the RHSM reconfigure script
# @api private
define foreman_proxy_content::rhsm_reconfigure_script($ca_cert, $server_ca_cert) {
```

templates/bootstrap_rpm/rhsm-katello-reconfigure.erb
```diff
@@ -0,0 +1,115 @@
+KATELLO_SERVER=<%= @rhsm_hostname %>
+KATELLO_SERVER_CA_CERT=<%= scope['certs::server_ca_name'] %>.pem
+KATELLO_DEFAULT_CA_CERT=<%= scope['certs::default_ca_name'] %>.pem
+KATELLO_CERT_DIR=<%= scope['certs::katello::rhsm_ca_dir'] %>
```
ekohl, line 4: I'm not a fan of this linking because it can easily break if we start to refactor things. Any thoughts on switching to EPP and explicitly passing it as context? It's more verbose, but we'll notice easily if it does break.

manifests/bootstrap_rpm.pp
ehelms: Sure.

manifests/bootstrap_rpm.pp
```diff
@@ -0,0 +1,66 @@
+# This class builds an RPM containing the bootstrap for a subscription-manager consumer
+# This file is placed in $katello_www_pub_dir.
+# @api private
+class foreman_proxy_content::bootstrap_rpm (
+  String $hostname = $facts['networking']['fqdn'],
```
ekohl, line 5:
```suggestion
  Stdlib::Fqdn $hostname = $facts['networking']['fqdn'],
```

manifests/bootstrap_rpm.pp
```diff
@@ -0,0 +1,66 @@
...
+
+  $candlepin_consumer_name = "katello-ca-consumer-${hostname}"
+  $candlepin_consumer_summary = "Subscription-manager consumer certificate for Katello instance ${hostname}"
+  $candlepin_consumer_description = 'Consumer certificate and post installation script that configures rhsm.'
+
+  include trusted_ca
+  trusted_ca::ca { 'katello_server-host-cert':
+    source  => $katello_server_ca_cert,
+    require => File[$katello_server_ca_cert],
+  }
```
ekohl, line 31: This is also a part I always wondered about: why does the server need to globally trust the server certificate? Don't we set the CA explicitly everywhere?

ehelms: Looks like, based on https://projects.theforeman.org/issues/13634, we could drop it for the reasons stated it was added. However, there is currently known code that does rely on this we would need to fix. I was discussing that last week with @jlsherrill

ekohl: Perhaps we can start with adding a class parameter `Boolean $add_ca_as_trusted_ca = true` so we can toggle it to false in nightly to see what breaks? Since it's a private class, users shouldn't see anything.

manifests/rhsm_reconfigure_script.pp
```diff
@@ -0,0 +1,65 @@
+# @summary Compose RHSM reconfigure script via concats
+# @api private
+define foreman_proxy_content::rhsm_reconfigure_script(
+  Stdlib::Absolutepath $ca_cert = undef,
+  Stdlib::Absolutepath $server_ca_cert = undef,
+  Stdlib::Absolutepath $rhsm_ca_dir = undef,
+  String $default_ca_name = undef,
+  String $server_ca_name = undef,
```
ekohl, line 8: In Puppet it's not strictly needed to have a default. Since the default is invalid for the type, you can leave off `= undef` to indicate they're required parameters. That was done in Puppet < 4 before there were data types but is still entirely valid.

manifests/bootstrap_rpm.pp
ehelms: I know at least Pulp communication will likely break due to how that is currently coded. Let me compromise with creating a dedicated issue to explore this (https://projects.theforeman.org/issues/30338) so that this code is still as clean as can be a copy from puppet-certs. Then it can be a targeted follow up along with the bootstrap RPM being re-factored to drop katello-cert-tools as a stand alone fix.

I do agree our code should be robust against the CA certificate being in the system wide truststore but are there other reasons that we shouldn't do this?

manifests/bootstrap_rpm.pp
```diff
@@ -0,0 +1,66 @@
...
+  # Placing the CA in the pub dir for trusting by a user in their browser
+  file { "${katello_www_pub_dir}/${server_ca_name}.crt":
+    ensure  => file,
+    source  => $katello_server_ca_cert,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0644',
+    require => File[$katello_server_ca_cert],
+  } ~>
+  foreman_proxy_content::rhsm_reconfigure_script { "${katello_www_pub_dir}/${katello_rhsm_setup_script}":
```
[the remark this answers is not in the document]
ehelms: Maybe? But I am trying not to over think redesigning this due to how much this is going to change when it gets re-factored away from katello-certs-tools under the hood and towards both a simpler puppet layer and using RPM directly. I know that's hard, but instead of big bang, I'm trying to find the baby steps I can take to allow easier review while maintaining functionality

manifests/bootstrap_rpm.pp
ekohl: The less invasive we are to the system, the easier. We can also drop the `trusted_ca` module from the installer. Right now I'm maintaining that in the Voxpupuli organization because we need it. That overhead might not be directly visible, but it's there.

I also don't know if there may be security policies on customer systems that would disallow this. Perhaps now they don't know so don't care, but might if they did.

Let's keep this and make it part of the "kill pulp 2"-PR

ehelms:
The most realistic deployment is that the reverse proxy + RHSM endpoint will always be collocated with Pulp deployment. Now, if we talk possible deployments, subscription-manager does support configuring its RHSM API endpoint differently than where the content lives. In other words, you can point at two different hostnames.

There is also the evolving of this general concept into being able to do simple registration (through a simple bash script most likely fetched from the Foreman itself) that should proxy through the external Foreman node (whether through smart proxy an Apache reverse proxy remains to be seen) back to the Foreman to isolate traffic similar to how we do RHSM today. If using a Red Hat based system, then RHSM end point should be deployed with local content.

@sbernard had a use case for pulling content through the reverse proxy from the main Foreman server which would imply RHSM endpoint but no Pulp -- https://github.com/theforeman/puppet-pulp/pull/372

So I'd say there are use cases for having pieces that we compose into the 90% case and could build into these use cases but that we don't over engineer it and risk missing the release deadlines so that can we make good, reasonable changes for most folks and set ourselves up to iterate quickly going forward.

ekohl:
It sounds to me like we could go with a single content plugin but have explicit capabilities that Katello should check for.

The first one is Pulp, which will bring in its own capabilities for each content type. If the Pulp capability is present, it should expose the setting `pulpcore_api_url` and `pulpcore_content_url` as valid URLs. Without the capability, the setting may be null and no assumptions about it should be made.

The other capability is RHSM, but I don't have a good grasp on what it'd need. It feels like today it's only used by Katello to tell the client where to get the registration RPM. Within this RPM the actual RHSM endpoint lives and the client uses this. It feels like we should expose both these URLs as settings. While Katello doesn't use the real RHSM endpoint today, it can be useful. Registration can be implemented with templates inside Foreman and a REX job can update the RHSM endpoints if it changed. All without having to guess. It'll also be possible to run RHSM on a different hostname than foreman-proxy.

Big question I have is whether it makes sense to combine both in a single Smart Proxy plugin or rather keep them as separate plugins and separate Smart Proxy Features.

Thinking about this, it feels like separate plugins is easier though.

I'm wondering what @jlsherrill thinks about this.

ehelms:
I think separate plugins makes sense and I would not invest in the RHSM one too much just yet as there is work on going to re-think the registration process and mechanisms (an RFC will be coming that goes into more detail from my understanding) that might influence the direction that plugin should go. 

Do you consider the, really good, design discussions here relevant to the specific change in this PR? Once this change is in, I would like to start re-factoring the underlying mechanisms and cleaning this up.

ehelms:
Now with acceptance tests

PR #273, opened 2020-07-10
Allow katello/certs 9.x

ehelms:


metadata.json
```diff
@@ -34,7 +34,7 @@
     },
     {
       "name": "katello/certs",
-      "version_requirement": ">= 6.0.0 < 9.0.0"
+      "version_requirement": ">= 9.0.0 < 10.0.0"
```
ekohl, line 37: In general we keep dependencies open. AFAIK this should still work with version 6
```suggestion
      "version_requirement": ">= 6.0.0 < 10.0.0"
```

ehelms: OK.. we can keep it open, but we do know that there are changes coming that will force this update anyway and that this module's dependence on puppet-certs should align with puppet-katello.

ekohl: Historically I've always been lazy in that: only do what's needed now. You don't know if you'll really need it. It's also a good thing to see that a PR also changes it the deps when it really needs to. Probably doesn't matter that much, but it's a habit I've built.

PR #274, opened 2020-07-14
Fixes #30363: Rewrite bootstrap creation to use RPM

ehelms:
Opening a draft PR to share my early direction and thought for how to convert the creation of the RPM to pure rpm commands and drop the use of katello-certs-tools for it.


lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+  end
+
+  def delete
+  end
+
+  def exists?
+    File.exist?(resource[:symlink]) if resource[:symlink]
+  end
+
+	private
```
ekohl, line 25: Indenting

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+
+  def write_specfile
+    File.open("#{spec_dir}/#{resource[:name]}.spec", 'w') do |file|
+      file.write(spec)
+    end
+  end
+
+  def release
+    if !latest_rpm.nil? && File.exist?(latest_rpm)
+      rel = `rpm -qp #{latest_rpm} --queryformat=%{release}`
```
ekohl, line 59: This is not how you should call commands. This should be done in the same way as `commands :rpmbuild`. Then you call it as `rpm('-qp', latest_rpm, '--queryformat=%{release}')`

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+      rel.to_i + 1
+    else
+      1
+    end
+  end
+
+  def build_rpm
+    command = ['rpmbuild', '-ba']
```
ekohl, line 67: You should also use `rpmbuild()` here to call the right command. Backticks don't use the proper error handling. You also get more debugging.

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+      install -Dp -m0755 %{SOURCE0} %{buildroot}%{_bindir}/katello-rhsm-consumer
+
+      %clean
+      rm -rf %{buildroot}
+
+      %post
+      /bin/bash /usr/bin/katello-rhsm-consumer
+
+      %postun
+      test -f /etc/rhsm/rhsm.conf.kat-backup && command cp /etc/rhsm/rhsm.conf.kat-backup /etc/rhsm/rhsm.conf
```
ekohl, line 133: Isn't this introducing a bug again?
https://github.com/theforeman/puppet-certs/commit/7bea39e9a03a5a34c675174f9e43b2bc44de51f0

manifests/bootstrap_rpm.pp
```diff
@@ -20,7 +20,7 @@
   $katello_rhsm_setup_script = 'katello-rhsm-consumer'
   $katello_rhsm_setup_script_location = "/usr/bin/${katello_rhsm_setup_script}"
 
-  $candlepin_consumer_name = "katello-ca-consumer-${rhsm_hostname}"
+  $bootstrap_rpm_name = "katello-ca-consumer-${rhsm_hostname}"
   $candlepin_consumer_summary = "Subscription-manager consumer certificate for Katello instance ${rhsm_hostname}"
   $candlepin_consumer_description = 'Consumer certificate and post installation script that configures rhsm.'
```
ekohl, line 25: These appear to be unused now. I do wonder if it doesn't make sense to keep those in.

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+  def spec_dir
+    mkdir('SPECS')
+  end
+
+  def rpm_dir
+    mkdir('RPMS')
+  end
+
+  def copy_sources
+    FileUtils.copy(resource[:script], "#{sources_dir}/#{File.basename(resource[:script])}")
```
ekohl, line 48:
```suggestion
    FileUtils.copy(resource[:script], File.join(sources_dir, File.basename(resource[:script]))
```

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+  end
+
+  def write_specfile
+    File.open("#{spec_dir}/#{resource[:name]}.spec", 'w') do |file|
```
ekohl, line 52:
```suggestion
    File.open(File.join(spec_dir, "#{resource[:name]}.spec"), 'w') do |file|
```

manifests/bootstrap_rpm.pp
```diff
@@ -56,14 +56,10 @@
...
-    description      => $candlepin_consumer_description,
-    files            => ["${katello_rhsm_setup_script_location}:755=${katello_www_pub_dir}/${katello_rhsm_setup_script}"],
-    bootstrap_script => "/bin/bash ${katello_rhsm_setup_script_location}",
-    postun_script    => file('certs/postun.sh'),
-    alias            => $candlepin_cert_rpm_alias_filename,
-    subscribe        => $server_ca,
+  bootstrap_rpm { $bootstrap_rpm_name:
+    ensure  => present,
+    script  => "${katello_www_pub_dir}/${katello_rhsm_setup_script}",
+    dest    => "${katello_www_pub_dir}",
```
ekohl, line 62:
```suggestion
    dest    => $katello_www_pub_dir,
```

lib/puppet/type/bootstrap_rpm.rb
```diff
@@ -0,0 +1,19 @@
+Puppet::Type.newtype(:bootstrap_rpm) do
```
ekohl, line 1: This should have autorequires (https://puppet.com/docs/puppet/latest/custom_types.html#tandp_automatic_relationships):
```ruby
autorequire(:file) do
  [self[:dest]]
end

autorequire(:concat) do
  [self[:script]]
end
```

manifests/bootstrap_rpm.pp
```diff
@@ -56,14 +56,10 @@
     rhsm_port       => $rhsm_port,
     rhsm_prefix     => $rhsm_prefix,
   } ~>
-  certs_bootstrap_rpm { $candlepin_consumer_name:
-    dir              => $katello_www_pub_dir,
-    summary          => $candlepin_consumer_summary,
-    description      => $candlepin_consumer_description,
-    files            => ["${katello_rhsm_setup_script_location}:755=${katello_www_pub_dir}/${katello_rhsm_setup_script}"],
-    bootstrap_script => "/bin/bash ${katello_rhsm_setup_script_location}",
-    postun_script    => file('certs/postun.sh'),
```
ekohl, line 65: This is ugly and I didn't spot it initially, but it was still reading the file from the old module. After this is merged, please open a PR to delete it.

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
ehelms: I tried that, whatever is underlying that functionality does not play nicely with rpmbuild. I had similar issues with shelling out in python to call rpmbuild in Obal. The macros don't get passed to rpmbuild correctly:

```
  Info: Foreman_proxy_content::Rhsm_reconfigure_script[/var/www/html/pub/katello-rhsm-consumer]: Scheduling refresh of Bootstrap_rpm[katello-ca-consumer-centos7-64.example.com]
  Error: Execution of '/usr/bin/rpmbuild -ba /tmp/d20200714-1701-os51rg/SPECS/katello-ca-consumer-centos7-64.example.com.spec --define "_topdir /tmp/d20200714-1701-os51rg" --define "name katello-ca-consumer-centos7-64.example.com" --define "release 1"' returned 1: error: Macro % has illegal name (%define)
  error: Macro % has illegal name (%define)
  error: Macro % has illegal name (%define)
  error: Macro % has illegal name (%define)
  error: Macro % has illegal name (%define)
  error: Macro % has illegal name (%define)
  error: Macro % has illegal name (%define)
  error: Too many levels of recursion in macro expansion. It is likely caused by recursive macro declaration.
```

ekohl: I think that's because you double quote it. I'll take a stab at this.

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+    link_rpm
+  ensure
+    FileUtils.remove_dir(base_dir) if File.directory?(base_dir)
+  end
+
+  def delete
+  end
+
+  def exists?
+    File.exist?(resource[:symlink]) if resource[:symlink]
```
ekohl, line 22: If symlink is not given this evaluates to nil and thus false. I think this is what you meant.
```suggestion
    !resource[:symlink] || File.exist?(resource[:symlink])
```

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
+require 'fileutils'
+
+Puppet::Type.type(:bootstrap_rpm).provide(:bootstrap_rpm) do
+
+  commands :rpmbuild => 'rpmbuild'
+
+  def create
+    base_dir
+    copy_sources
```
ekohl, line 9:
```suggestion
    sources_dir
    copy_sources
```

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
ekohl: This works for me:
```ruby
  def build_rpm
    rpmbuild('-ba', File.join(spec_dir, "#{resource[:name]}.spec"),
             '--define', "_topdir #{base_dir}",
             '--define', "name #{resource[:name]}",
             '--define', "release #{release}")
  end 
```

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
ekohl: Hmm, this should not be needed but it's consistently failing for me in EL8. Perhaps it's calling something after the base_dir has been removed?

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+    write_specfile
+    build_rpm
+    copy_rpm
+    link_rpm
+  ensure
+    FileUtils.remove_dir(base_dir) if File.directory?(base_dir)
```
ekohl, line 15: To get around the `SOURCES` issue I had to use:
```suggestion
    FileUtils.remove_dir(base_dir) if File.directory?(base_dir)
    @base_dir = nil
```

This does suggest `create` is called more than once. Likely that `exists?` is incorrect.

ehelms: Yes, its the refresh triggering it as the full idempotency of this is not accounted for yet. I wanted to get draft feedback on the overall design approach before solving the complex issue of detecting when a new RPM is needed.

manifests/bootstrap_rpm.pp
```diff
@@ -19,51 +16,49 @@
...
     require => File[$katello_server_ca_cert],
   }
 
+  package { 'rpmbuild':
+    ensure => installed,
+  }
+
   include apache
-  file { $katello_www_pub_dir:
+  file { $rpm_serve_dir:
```
ehelms, line 34: Does not have to be this PR, but @ekohl how do you envision we should drop management of this here in favor of elevating it to other places we manage Apache?

ekohl: I noticed the exact same thing yesterday. Not entirely sure yet. I'm wondering if we will continue to use an RPM to register or fully move to registration templates. That would remove the need entirely and then there's no reason to make the integration cleaner.

ehelms: The RPM will exist for at least 2-4 more releases as we will want to give users time to adapt their host workflows to the new method before removing the old.

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+  def build_rpm
+    output = rpmbuild(
+      '-ba',
+      File.join(spec_dir, "#{resource[:name]}.spec"),
+      '--define', "_topdir #{base_dir}",
+      '--define', "name #{resource[:name]}",
+      '--define', "release #{release}"
+    )
+
+    Puppet.debug(output)
```
ekohl, line 72: I wonder if Puppet can't do this by default when it calls commands. Perhaps if you use `--trace`?

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+    end
+  end
+
+  def link_rpm
+    FileUtils.ln_s(latest_rpm, resource[:symlink], force: true) if resource[:symlink]
+  end
+
+  def latest_rpm
+    latest = Dir.glob("#{resource[:dest]}/*.rpm").sort.reverse
+    false if latest.nil?
```
ekohl, line 87: Did you intend to return?
```suggestion
    return false if latest.nil?
```

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+      FileUtils.copy(file, resource[:dest])
+    end
+  end
+
+  def link_rpm
+    FileUtils.ln_s(latest_rpm, resource[:symlink], force: true) if resource[:symlink]
+  end
+
+  def latest_rpm
+    latest = Dir.glob("#{resource[:dest]}/*.rpm").sort.reverse
```
ekohl, line 86: Alphabetical sorting will fail on `-10` vs `-2`.

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+      rm -rf %{buildroot}
+
+      install -d -m0755 %{buildroot}%{_bindir}
+      install -Dp -m0755 %{SOURCE0} %{buildroot}%{_bindir}/katello-rhsm-consumer
+
+      %clean
+      rm -rf %{buildroot}
+
+      %post
+      /bin/bash /usr/bin/katello-rhsm-consumer
```
ekohl, line 128: It's executable with a shebang. Is `/bin/bash` really needed?

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+
+  def mkdir(dir)
+    path = File.join(base_dir, dir)
+    Dir.mkdir(path) unless File.exist?(path)
+    path
+  end
+
+  def spec
+    <<~HEREDOC
+      Name:      %{name}
```
ekohl, line 99: Isn't the Package name normally prefixed? Or the summary will be a bit odd

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,139 @@
...
+      Consumer certificate and post installation script that configures rhsm.
+
+      %prep
+
+      %build
+
+      %install
+      rm -rf %{buildroot}
+
+      install -d -m0755 %{buildroot}%{_bindir}
```
ekohl, line 121: This is redundant since the next line uses `install -D`

manifests/bootstrap_rpm.pp
```diff
@@ -19,51 +16,48 @@
...
     ensure => directory,
     owner  => 'apache',
     group  => 'apache',
     mode   => '0755',
```
ekohl, line 38: I was wondering why Apache needs to write to this directory. Given it was there before, I don't mind keeping it.

manifests/bootstrap_rpm.pp
ekohl: I also want to fix it, but let's keep it for the next iteration.

templates/bootstrap_rpm/rhsm-katello-reconfigure.erb
```diff
@@ -1,10 +1,10 @@
 KATELLO_SERVER=<%= @rhsm_hostname %>
 KATELLO_SERVER_CA_CERT=<%= @server_ca_name %>.pem
 KATELLO_DEFAULT_CA_CERT=<%= @default_ca_name %>.pem
-KATELLO_CERT_DIR=<%= @rhsm_ca_dir %>
 PORT=<%= @rhsm_port %>
 
-PREFIX=<%= @rhsm_prefix %>
+CERT_DIR='/etc/rhsm/ca'
+PREFIX=/rhsm
```
ekohl, line 7: I actually would like to keep this a variable for now. The `/rhsm` will be used in the Apache config as well and when it's a variable, you can see it's related. Even though it's unlikely to change. This is because Puppet doesn't really have constants, but constants(/variables) are a really good way of indicating related systems.

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
ehelms: I don't know. How does that work through the installer? This felt like a nice easy option.

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
ehelms: Prefixed?

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
ekohl: Not sure. It just feels that this is hard to trace down. Just some debug output without knowing where it comes from.

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
ekohl: It looks like the name is set to the resource name. That's katello-ca-consumer-${rhsm_hostname}. Then in `%summary` it says `Katello instance %{name}`. That looks odd.

templates/bootstrap_rpm/rhsm-katello-reconfigure.erb
ehelms: The /rhsm is not used in the Apache config today:

```
# grep -r rhsm /etc/httpd/
```

As the reverse proxy proxies all traffic back to Foreman and Katello defines a `/rhsm` API. So there is nothing to relate it back to it from what I can tell. This just "is".

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
ehelms: I adjusted the summary to be more descriptive but generic of the name

ehelms:
@ekohl Is there a way in the acceptance tests to say "run all these acceptance tests and *then* run another manifest application and perform additional verifications" (in that strict order) ?

ekohl:
Summarizing the IRC discussion: https://relishapp.com/rspec/rspec-core/docs/configuration/overriding-global-ordering#running-a-specific-example-group-in-order or avoid the `it_behaves_like` helper and write it up manually within the same context since `it` blocks do appear to be deterministic. Effectively it is part of the same `describe` block so that would be the correct way IMHO.

ehelms:
This contains a refactor of the bootstrap handling code to use native rpm and moves the rhsm reconfigure script into a custom provider to have better vision of the script as a whole and control of it. To aid with the code changes, I've made the bootstrap RPM creation more idempotent and added additional tests to ensure conditions such as CA certificate changing or an RHSM configuration option (i.e. port) leads to a new RPM being created. The code got a bit more complex, but detecting that an RPM had "changed" took some work.

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,201 @@
...
+    end
+  ensure
+    FileUtils.remove_dir(base_dir) if File.directory?(base_dir)
+    @base_dir = nil
+  end
+
+  def exists?
+    copy_sources
+    write_specfile
+    build_rpm
```
ekohl, line 28: Building an RPM on every single run feels a bit heavy. Isn't there a way to avoid this? For example, you can add a `Provides: katello-ca-cert(hash) = $HASH` statement and similar for the script. That way you can query it with `rpm`

ehelms:
It's quite fast given how simple it is. The two concerns that we have to
consider are: katello-rhsm-changes and spec file changes.

I could store both hashes on the RPM via provides?

ehelms:
@ekohl Further thoughts here ?

lib/puppet/type/rhsm_reconfigure_script.rb
```diff
@@ -0,0 +1,31 @@
+Puppet::Type.newtype(:rhsm_reconfigure_script) do
+  desc 'Creates script to reconfigure RHSM to point at a server'
+
+  ensurable
+
+  newparam(:name, :namevar => true)
```
ekohl, line 6: The name is really the path, right? You don't need the namevar to be `name`: https://puppet.com/docs/puppet/5.5/custom_types.html#namevar

lib/puppet/provider/rhsm_reconfigure_script/rhsm_reconfigure_script.rb
```diff
@@ -0,0 +1,151 @@
...
+      KATELLO_DEFAULT_CA_CERT=<%= resource[:default_ca_name] %>.pem
+
+      CERT_DIR=/etc/rhsm/ca
+      PREFIX=/rhsm
+      CFG=/etc/rhsm/rhsm.conf
+      CFG_BACKUP=$CFG.kat-backup
+      CA_TRUST_ANCHORS=/etc/pki/ca-trust/source/anchors
+
+      read -r -d '' KATELLO_DEFAULT_CA_DATA << EOM || true
+      <%= File.read(resource[:default_ca_cert]) %>
```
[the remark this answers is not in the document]
ehelms: I thought I always understood it that providers were executed on clients as they acted like any other resource (e.g. file, service, exec)

lib/puppet/provider/rhsm_reconfigure_script/rhsm_reconfigure_script.rb
```diff
@@ -0,0 +1,151 @@
...
+    ::ERB.new(script_template).result(binding)
+  end
+
+  def script_template
+    <<~'HEREDOC'
+      #!/bin/bash
+
+      set -e
+
+      KATELLO_SERVER=<%= resource[:rhsm_hostname] %>
```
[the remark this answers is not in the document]
ehelms: Didn't occur to me to try at the time, I was just translating the original over for the most part. I can give regular string interpolation a try igven I think the escaping and formatting are driven by the HEREDOC.

lib/puppet/provider/rhsm_reconfigure_script/rhsm_reconfigure_script.rb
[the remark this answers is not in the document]
ekohl: I think you may be right indeed. I still feel that a helper method with a comment can explain why this is a type + provider rather than a file + template. Normally I'm minimalistic in my comments but when something is unusual it helps to highlight what part is unusual.

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,201 @@
...
+
+    rpms = rpms.collect do |rpm|
+      version = rpm('-qp', rpm, "--queryformat=%{version}")
+      release = rpm('-qp', rpm, "--queryformat=%{release}")
+
+      {'release' => release, 'rpm' => rpm}
+    end
+
+    rpms = rpms.sort { |a,b| a['release'].to_i <=> b['release'].to_i }
+    rpms.last['rpm']
```
ekohl, line 151: Isn't this equivalent?
```suggestion
    rpms.max_by { |rpm| rpm['release'].to_i }['rpm']
```

spec/acceptance/bootstrap_rpm_spec.rb
```diff
@@ -58,6 +63,92 @@
...
+        ensure => latest,
+        source => "/var/www/html/pub/katello-ca-consumer-latest.noarch.rpm",
+        require => Class['foreman_proxy_content::bootstrap_rpm'],
+      }
+      EOS
+    end
+
+    it 'applies with no errors' do
+      apply_manifest(pp, catch_failures: true)
+    end
```
ekohl, line 125: Can we also check for idempotency?

spec/acceptance/bootstrap_rpm_spec.rb
```diff
@@ -58,6 +63,92 @@
...
```
ekohl, line 89: Can we also check for idempotency?

spec/acceptance/bootstrap_rpm_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper_acceptance'
 
-describe 'bootstrap_rpm' do
+describe 'bootstrap_rpm', :order => :defined do
```
ekohl, line 3: Should this remove all RPMs in `/var/www/html/pub/` before we start so you can run the test multiple times? If you now run rspec multiple times then it'll keep increasing the number.

lib/puppet/provider/rhsm_reconfigure_script/rhsm_reconfigure_script.rb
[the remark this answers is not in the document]
ehelms: This turned out ot be a combo of needing to be able to use single-quoted HEREDOC for formatting and escaping and ERB for the "interpolation".

spec/acceptance/bootstrap_rpm_spec.rb
```diff
@@ -1,6 +1,10 @@
...
+
+  before(:suite) do
+    `rm -rf /var/www/html/pub/*rpm`
+  end
```
ehelms, line 7: Not the cleanest but does what you wanted I think.

spec/acceptance/bootstrap_rpm_spec.rb
```diff
@@ -1,6 +1,10 @@
 require 'spec_helper_acceptance'
 
-describe 'bootstrap_rpm' do
+describe 'bootstrap_rpm', :order => :defined do
+
+  before(:suite) do
+    `rm -rf /var/www/html/pub/*rpm`
```
ekohl, line 6: This executes locally, not on the remote server. You can verify this by running it twice and you'll see it fails:
```
BEAKER_destroy=no be rspec spec/acceptance/bootstrap_rpm_spec.rb
BEAKER_provision=no BEAKER_destroy=no be rspec spec/acceptance/bootstrap_rpm_spec.rb
```

The second run reuses the machine and its state, which makes iterating on a change much faster.

This makes it pass for me:

```suggestion
  before(:all) do
    on hosts, 'rm -rf /var/www/html/pub/*rpm'
```

lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -0,0 +1,200 @@
...
+    return false if rpms.empty?
+
+    rpms = rpms.collect do |rpm|
+      version = rpm('-qp', rpm, "--queryformat=%{version}")
+      release = rpm('-qp', rpm, "--queryformat=%{release}")
+
+      {'release' => release, 'rpm' => rpm}
+    end
+
+    rpms.max_by { |rpm| rpm['release'].to_i }['rpm']
```
ekohl, line 150: Btw, it would be more efficient to only query the version once you need it:
```suggestion
    latest = rpms.max_by { |name| rpm('-qp', name, "--queryformat=%{release}") }
    rpm('-qp', latest, "--queryformat=%{version}")
```

This may also fix some issues with masking. looks like you use `rpm` both as a local variable and as a method.

Code is untested though, so be aware of that :)

manifests/bootstrap_rpm.pp
```diff
@@ -19,51 +16,52 @@
...
-    alias            => $candlepin_cert_rpm_alias_filename,
-    subscribe        => $server_ca,
+  }
+
+  bootstrap_rpm { $bootstrap_rpm_name:
+    ensure  => present,
+    script  => "${rpm_serve_dir}/${katello_rhsm_setup_script}",
+    dest    => $rpm_serve_dir,
+    symlink => "${rpm_serve_dir}/${candlepin_cert_rpm_alias_filename}",
+    require => Rhsm_reconfigure_script["${rpm_serve_dir}/${katello_rhsm_setup_script}"],
```
ekohl, line 65: This suggests it could have an autorequire in the type:
```ruby
autorequire(:rhsm_reconfigure_script) do
  [self[:script]]
end
```

ekohl (approved):
Small nits, but overall :+1:

manifests/bootstrap_rpm.pp
ehelms: Already does -- https://github.com/theforeman/puppet-foreman_proxy_content/pull/274/files#diff-b3f88bbcbe8556735e4d1d0ec3017af9R26

ekohl: In that case this line is redundant and can be dropped.

PR #275, opened 2020-07-17
Refs #30423 - Expose pulpcore_media_root in puppet-FPC

wbclark:
Requires: https://github.com/theforeman/puppet-pulpcore/pull/113


wbclark:
> Why do we need to expose this? I don't think this is something we want to support. If they really need to, they can change it via hiera.

This is why: https://github.com/theforeman/foreman-installer/pull/543

ekohl:
Closing in favor of the alternative approach

PR #278, opened 2020-08-27
Fixes #30716: Ensure /pub on foreman proxy can be browsed by default

ehelms:
Fixes regression introduced by f6790c4

(cherry picked from commit 8cfa178046a4dcd80f9e21c2f517f2cc75f67bcf)


ekohl (approved):
I changed the base branch from master to 11.1-stable. That is what you intended, right?

ehelms:
Ahh yep, thanks for catching that.

PR #281, opened 2020-09-04
Fixes #30780 - support pulp_container 2.0 registry

jlsherrill:


jlsherrill:
/pulpcore_registry/  is now proxying to the pulpcore-api app.  Calls to that registry will redirect to /pulp/container/.

This isn't fully working yet (although i think this is part of the solution), so leaving it in draft for now

jlsherrill:
https://pulp.plan.io/issues/7462  was the last piece of the puzzle, so this is now ready to go (in anticipation of that change).  This is a backwards incompatible change with pulpcore 3.5 or earlier (if that matters)

templates/pulpcore-docker-apache.conf.erb
```diff
@@ -2,5 +2,8 @@
   SSLRequire %{SSL_CLIENT_S_DN_CN} eq "admin"
```
ekohl, line 2: Is this still needed? Doesn't the API handle authentication now? Should we pass the the variable to the API backend?

jlsherrill:
Yes, we could just pass the variable through, similar to /pulp/api/v3, good
point, will update!

templates/pulpcore-docker-apache.conf.erb
jlsherrill: added!

templates/pulpcore-docker-apache.conf.erb
```diff
@@ -1,6 +1,10 @@
...
+  RequestHeader set REMOTE_USER ""
+  RequestHeader set REMOTE_USER "%{SSL_CLIENT_S_DN_CN}s" env=SSL_CLIENT_S_DN
 </Location>
 
-ProxyPass /pulpcore_registry/v2/ http://<%= scope['pulpcore::content_host'] %>:<%= scope['pulpcore::content_port']%>/v2/
-ProxyPassReverse /pulpcore_registry/v2/ http://<%= scope['pulpcore::content_host'] %>:<%= scope['pulpcore::content_port'] %>/v2/
+ProxyPass /pulpcore_registry/v2/ http://<%= scope['pulpcore::api_host'] %>:<%= scope['pulpcore::api_port']%>/v2/
+ProxyPassReverse /pulpcore_registry/v2/ http://<%= scope['pulpcore::api_host'] %>:<%= scope['pulpcore::api_port'] %>/v2/
+
+ProxyPass /pulp/container/ http://<%= scope['pulpcore::content_host'] %>:<%= scope['pulpcore::content_port']%>/pulp/container/
```
ekohl, line 9: How about this location? Should this also be receiving REMOTE_USER?

ekohl (commented):
Nice. Looks like this unifies all plugins in the same pattern.

templates/pulpcore-docker-apache.conf.erb
jlsherrill: nope, it doesn't.  all the auth to that endpoint is handled by the generated tokens passed via the redirect from the /v2/ endpoint in the api app

jlsherrill: actually, it looks like a crazy request like this:

 curl https://`hostname`/pulp/container/../api/v3/status/
works, but i can't get something like this:

curl "https://`hostname`/pulp/container/../api/v3/tasks/" -H 'REMOTE_USER: admin'

to work.  it still complains about auth.     It seems like apache is doing the right thing here and recognizing that '../api/v3/tasks' actually matches the '/pulp/api/v3/' Location.  If i comment out the "RequestHeader set REMOTE_USER " line from the /pulp/api/v3/ location, it then works. 

So yes, this is good as is

ekohl: Ok. In https://github.com/theforeman/puppet-pulpcore/pull/105 I'll probably revisit this and see how it turns out.

ekohl (approved):

PR #282, opened 2020-09-11
deploy pub_dir also when not deploying pulp2

evgeni:
we need /pub for the bootrap rpm/certs in any case

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -267,8 +267,9 @@ class foreman_proxy_content (
       require      => Class['certs::apache'],
     }
 
-    include foreman_proxy_content::pub_dir
   }
+
+  include foreman_proxy_content::pub_dir
 
   if $pulp {
     include apache
@@ -325,10 +326,8 @@ class foreman_proxy_content (
       worker_timeout         => $pulp_worker_timeout,
     }
 
-    $pub_dir_options = '+FollowSymLinks +Indexes'
-
     pulp::apache::fragment{'gpg_key_proxy':
-      ssl_content => template('foreman_proxy_content/_pulp_gpg_proxy.erb', 'foreman_proxy_content/httpd_pub.erb'),
+      ssl_content => template('foreman_proxy_content/_pulp_gpg_proxy.erb'),
     }
   }
 
--- a/manifests/pub_dir.pp
+++ b/manifests/pub_dir.pp
@@ -23,7 +23,7 @@ class foreman_proxy_content::pub_dir (
       content     => template('foreman_proxy_content/httpd_pub.erb'),
       ssl_content => template('foreman_proxy_content/httpd_pub.erb'),
     }
-  } elsif ($foreman_proxy_content::pulp) {
+  } elsif $foreman_proxy_content::pulp {
     pulp::apache::fragment { 'pub_dir':
       ssl_content => template('foreman_proxy_content/httpd_pub.erb'),
     }
--- a/manifests/pub_dir.pp
+++ b/manifests/pub_dir.pp
@@ -16,13 +16,34 @@ class foreman_proxy_content::pub_dir (
 
   ensure_packages('katello-client-bootstrap')
 
-  apache::vhost { 'foreman_proxy_content':
-    servername          => $servername,
-    port                => 80,
-    priority            => '05',
-    docroot             => '/var/www/html',
-    options             => ['SymLinksIfOwnerMatch'],
-    additional_includes => ["${apache::confd_dir}/pulp-vhosts80/*.conf"],
-    custom_fragment     => template('foreman_proxy_content/httpd_pub.erb'),
+  if ($foreman_proxy_content::pulpcore or $foreman_proxy_content::pulp_master) {
+    include foreman::config::apache
+
+    foreman::config::apache::fragment { 'pub_dir':
+      content     => template('foreman_proxy_content/httpd_pub.erb'),
+      ssl_content => template('foreman_proxy_content/httpd_pub.erb'),
+    }
+  } elsif ($foreman_proxy_content::pulp) {
+    pulp::apache::fragment { 'pub_dir':
+      ssl_content => template('foreman_proxy_content/httpd_pub.erb'),
+    }
+
+    file { "${apache::confd_dir}/pulp-vhosts80/pub_dir.conf":
+      ensure  => file,
+      content => template('foreman_proxy_content/httpd_pub.erb'),
+      owner   => 'apache',
+      group   => 'apache',
+      mode    => '0600',
+      notify  => Service['httpd'],
+    }
+  } else {
+    apache::vhost { 'foreman_proxy_content':
+      servername      => $servername,
+      port            => 80,
+      priority        => '05',
+      docroot         => '/var/www/html',
+      options         => ['SymLinksIfOwnerMatch'],
+      custom_fragment => template('foreman_proxy_content/httpd_pub.erb'),
+    }
   }
 }
--- a/spec/classes/foreman_proxy_content_spec.rb
+++ b/spec/classes/foreman_proxy_content_spec.rb
@@ -18,6 +18,7 @@ describe 'foreman_proxy_content' do
 
         it { is_expected.to compile.with_all_deps }
         it { is_expected.to contain_package('katello-debug') }
+        it { is_expected.to contain_class('foreman_proxy_content::pub_dir') }
       end
 
       context 'with pulp', if: facts[:operatingsystemmajrelease] == '7' do
@@ -69,6 +70,7 @@ describe 'foreman_proxy_content' do
         end
 
         it { is_expected.to compile.with_all_deps }
+        it { is_expected.to contain_class('foreman_proxy_content::pub_dir') }
         it { is_expected.to contain_class('pulpcore').with(manage_apache: false).that_comes_before('Class[foreman_proxy::plugin::pulp]') }
 
         it do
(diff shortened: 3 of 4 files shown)


manifests/pub_dir.pp
```diff
@@ -16,13 +16,19 @@
 
   ensure_packages('katello-client-bootstrap')
 
+  if ($foreman_proxy_content::pulp_master or $foreman_proxy_content::pulp) {
+    $additional_includes = ["${apache::confd_dir}/pulp-vhosts80/*.conf"]
+  } else {
+    $additional_includes = []
+  }
+
   apache::vhost { 'foreman_proxy_content':
```
ekohl, line 25: I often wondered about this and it doesn't really make sense to me.

On a regular Katello install, there's Foreman listening on port 80 with the same servername (as @evgeni kindly provided):
```
[root@centos7-katello-3-16 httpd]# httpd -S
VirtualHost configuration:
*:5000                 centos7-katello-3-16.yatsu.example.com (/etc/httpd/conf.d/03-crane.conf:6)
*:443                  centos7-katello-3-16.yatsu.example.com (/etc/httpd/conf.d/05-foreman-ssl.conf:6)
*:80                   is a NameVirtualHost
         default server centos7-katello-3-16.yatsu.example.com (/etc/httpd/conf.d/05-foreman.conf:6)
         port 80 namevhost centos7-katello-3-16.yatsu.example.com (/etc/httpd/conf.d/05-foreman.conf:6)
                 alias foreman
         port 80 namevhost centos7-katello-3-16.yatsu.example.com (/etc/httpd/conf.d/05-foreman_proxy_content.conf:6)
```

There `/pub` is provided by https://github.com/theforeman/puppet-katello/blob/7200bf3152230ac938fd9b088c005c1f13ed38c9/templates/pulp-apache.conf.erb#L5-L11

So on a Katello install, this vhost is redundant and only confusing.

I don't have the data for a content proxy, but there we have Pulp 2 running instead of a Foreman and this code is triggered:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/bd6654b038c6dc41518fa59469808a4a28a88a73/templates/httpd_pub.erb#L1-L9

So effectively, this vhost is only relevant on a machine where there is no Foreman and no Pulp 2.

In https://github.com/theforeman/puppet-foreman_proxy_content/ I put `/pub` in the Pulp 3 vhost.

So effectively I think what you can do is leave out `additional_includes` altogether. Then you can move `include foreman_proxy_content::pub_dir` to an else branch in `init.pp`.

The only thing I wonder about is the `ensure_packages('katello-client-bootstrap')` but I wonder if that's really needed.

ekohl: In https://github.com/theforeman/puppet-foreman_proxy_content/pull/283 I implemented that.

manifests/pub_dir.pp
```diff
@@ -12,17 +12,33 @@
...
-    options             => ['SymLinksIfOwnerMatch'],
-    additional_includes => ["${apache::confd_dir}/pulp-vhosts80/*.conf"],
-    custom_fragment     => template('foreman_proxy_content/httpd_pub.erb'),
+  if ($foreman_proxy_content::pulp_master or $foreman_proxy_content::pulp) {
+    $additional_includes = ["${apache::confd_dir}/pulp-vhosts80/*.conf"]
+  } else {
+    $additional_includes = []
+  }
+
+  if ($foreman_proxy_content::pulpcore or $foreman_proxy_content::pulp_master) {
```
evgeni, line 24: Is there a better way to detect "this is katello, not proxy"?

evgeni:
On a Proxy, this still results in two vhosts on :80:

```
[root@centos7-foreman-proxy-nightly conf.d]# httpd -S
VirtualHost configuration:
*:5000                 centos7-foreman-proxy-nightly.yatsu.example.com (/etc/httpd/conf.d/03-crane.conf:6)
*:80                   is a NameVirtualHost
         default server centos7-foreman-proxy-nightly.yatsu.example.com (/etc/httpd/conf.d/05-foreman_proxy_content.conf:6)
         port 80 namevhost centos7-foreman-proxy-nightly.yatsu.example.com (/etc/httpd/conf.d/05-foreman_proxy_content.conf:6)
         port 80 namevhost centos7-foreman-proxy-nightly.yatsu.example.com (/etc/httpd/conf.d/05-pulp-http.conf:6)
                 alias centos7-foreman-proxy-nightly
*:443                  centos7-foreman-proxy-nightly.yatsu.example.com (/etc/httpd/conf.d/05-pulp-https.conf:6)
*:8443                 centos7-foreman-proxy-nightly.yatsu.example.com (/etc/httpd/conf.d/28-katello-reverse-proxy.conf:6)
ServerRoot: "/etc/httpd"
Main DocumentRoot: "/etc/httpd/htdocs"
Main ErrorLog: "/var/log/httpd/error_log"
Mutex ssl-stapling: using_defaults
Mutex proxy: using_defaults
Mutex ssl-cache: using_defaults
Mutex default: dir="/run/httpd/" mechanism=default 
Mutex mpm-accept: using_defaults
Mutex authdigest-opaque: using_defaults
Mutex rewrite-map: using_defaults
Mutex authdigest-client: using_defaults
PidFile: "/etc/httpd/run/httpd.pid"
Define: _RH_HAS_HTTPPROTOCOLOPTIONS
Define: DUMP_VHOSTS
Define: DUMP_RUN_CFG
User: name="apache" id=48
Group: name="apache" id=48
```

evgeni:
and it's fine on katello with https://github.com/theforeman/puppet-katello/pull/362

```
[root@centos8-katello-nightly ~]# httpd -S
VirtualHost configuration:
*:443                  centos8-katello-nightly.yatsu.example.com (/etc/httpd/conf.d/05-foreman-ssl.conf:6)
*:80                   centos8-katello-nightly.yatsu.example.com (/etc/httpd/conf.d/05-foreman.conf:6)
```

manifests/pub_dir.pp
```diff
@@ -12,17 +12,38 @@
...
+    include foreman::config::apache
+
+    foreman::config::apache::fragment { 'pub_dir':
+      content     => template('foreman_proxy_content/httpd_pub.erb'),
+      ssl_content => template('foreman_proxy_content/httpd_pub.erb'),
+    }
+  } elsif ($foreman_proxy_content::pulp) {
+    pulp::apache::fragment { 'pub_dir':
+      #content => template('foreman_proxy_content/httpd_pub.erb'),
```
evgeni, line 33: `pulp::apache::fragment` has no `content`, how do I get content into the 80 vhost?!

evgeni: wow, really, :shower:

evgeni:
I think I understand why Ewoud wanted to re-do all of this.

evgeni:
So this is now slightly more than "deploy /pub without pulp2", but at least we now have a central place to break /pub ;)

manifests/init.pp
```diff
@@ -267,9 +267,10 @@
       require      => Class['certs::apache'],
     }
 
-    include foreman_proxy_content::pub_dir
   }
 
+  include foreman_proxy_content::pub_dir
```
ehelms, line 272: This is now deployed at all times with FPC right? If I read this correctly. That follows our assumption that if you are deploying an FPC you intend to have a pub dir.

manifests/pub_dir.pp
ehelms: Not today.

ehelms (approved):
Tested, and for the current state this looks to be a good solution for handling this and moving forward the no Pulp 2 case that happens with EL8. I'll wait for @ekohl to weigh in as well

manifests/init.pp
evgeni: Yes, this is now "you're deploying f-p-c? you got a pubdir"

the only thing I'm curious about is whether `pub_dir.pp` actually needs that last else branch, aka: is fpc ever deployed without *any* pulp?

ekohl: In c363e7c0b068e7d2d06c5c0698c98f545c1c1f31 it was extracted, but it looks like 3d542ff831ebb167267de0e62eaa2a4fb63646f3 already introduced it. Back then it appeared to have been to expose the Pulp HTTP parts but that was superseded. So I think there isn't really a case where it makes sense.

manifests/pub_dir.pp
```diff
@@ -16,13 +16,34 @@
...
-    additional_includes => ["${apache::confd_dir}/pulp-vhosts80/*.conf"],
-    custom_fragment     => template('foreman_proxy_content/httpd_pub.erb'),
+  if ($foreman_proxy_content::pulpcore or $foreman_proxy_content::pulp_master) {
+    include foreman::config::apache
+
+    foreman::config::apache::fragment { 'pub_dir':
+      content     => template('foreman_proxy_content/httpd_pub.erb'),
+      ssl_content => template('foreman_proxy_content/httpd_pub.erb'),
+    }
+  } elsif ($foreman_proxy_content::pulp) {
```
ekohl, line 26: Nit: I always leave out the braces on single line statements.
```suggestion
  } elsif $foreman_proxy_content::pulp {
```

ekohl:
Does this also need a corresponding puppet-katello PR to avoid deploying it twice?

ehelms:
It has one referenced a few comments above -- https://github.com/theforeman/puppet-katello/pull/362

manifests/init.pp
ekohl: From a logic grouping perspective this is very closely related to our [bootstrap_rpm](https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/bootstrap_rpm.pp). I wonder if that should actually include the pub dir code. That also feels unnatural.

A more logical thing might be to include bootstrap_rpm from pub_dir or even merge the two classes.

I'd be ok with doing that in a follow up PR.

PR #283, opened 2020-09-11
Set up a pub_dir vhost on non-Pulp

ekohl:
In puppet-katello the /pub alias is set up as part of the foreman vhost on port 80 via katello::pulp. If Pulp 2 is managed by this module, the /pub alias is set up as part of the Pulp 2 HTTP vhost. That means it's only needed if Pulp 2 is not present at all, the foreman_proxy_content vhost is needed. It does mean that katello-client-bootstrap needs to be ensured in the case where Pulp 2 is present.

This also means the additional pulp-vhost80 includes are not needed.

I haven't verified this yet, but it's based on https://github.com/theforeman/puppet-foreman_proxy_content/pull/282/files#r486961343.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -267,6 +267,8 @@ class foreman_proxy_content (
       require      => Class['certs::apache'],
     }
 
+    ensure_packages('katello-client-bootstrap')
+  } else {
     include foreman_proxy_content::pub_dir
   }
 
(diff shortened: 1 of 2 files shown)


evgeni:
As discussed last week, we would like to move the /pub dir handling to one place, and this place being this module.

I tried to implement that in #282, but didn't touch the additional includes (yet).

PR #288, opened 2020-10-22
Bump pulpcore dep to 1.3.0 for Katello 3.17

chris1984:


ekohl:
This isn't really needed. The correct place to fix it is here:
 https://github.com/theforeman/foreman-installer/blob/42de236b41745bf0ce80fa9f5ec5a3ce0ee86206/Puppetfile.lock#L118
 https://github.com/theforeman/foreman-installer/blob/42de236b41745bf0ce80fa9f5ec5a3ce0ee86206/Puppetfile#L10

PR #292, opened 2020-11-05
Proxy pulp/deb to pulpcore support

jlsherrill:


jlsherrill:
I think i did this right?

manifests/init.pp
```diff
@@ -40,6 +40,8 @@
 #
 # $pulp_ca_cert::                       Absolute path to PEM encoded CA certificate file, used by Pulp to validate the identity of the broker using SSL.
 #
+# $proxy_pulp_deb_to_pulpcore::        Proxy /pulp/deb to Pulpcore at /pulp/content
```
ekohl, line 43: Alignment looks off. Not sure if it's Github rendering it off
```suggestion
# $proxy_pulp_deb_to_pulpcore::         Proxy /pulp/deb to Pulpcore at /pulp/content
```

manifests/init.pp
```diff
@@ -416,6 +420,9 @@
     class { 'pulpcore::plugin::rpm':
       use_pulp2_content_route => $proxy_pulp_yum_to_pulpcore,
     }
+    class { 'pulpcore::plugin::deb':
+      use_pulp2_content_route => $proxy_pulp_deb_to_pulpcore,
```
ekohl, line 424: This requires a corresponding PR to https://github.com/theforeman/puppet-pulpcore/blob/master/manifests/plugin/deb.pp to allow this.

jlsherrill: ahh thanks!

ekohl:
This depends on https://github.com/theforeman/puppet-pulpcore/pull/145. We should also take note to increase the minimum version to one that includes the change once merged.

wbclark (commented):
@jlsherrill could you update metadata.json here to require puppet-pulpcore version 2.1 where support for the pulp2 content route is available in the deb plugin ?

ekohl (approved):

jlsherrill:
@wbclark does this look okay now?

wbclark (approved):
Yes, it looks good to me. Thanks @jlsherrill !

PR #294, opened 2020-11-16
Fix missing variable scoping on reverse_proxy and add tests for pulpcore

ehelms:


wbclark (commented):
I would suggest a Redmine to link this to the previous PR, but the previous PR didn't have one. In that case how about referencing the previous commit in the commit message here?

ehelms:
> I would suggest a Redmine to link this to the previous PR, but the previous PR didn't have one. In that case how about referencing the previous commit in the commit message here?

Done. You can see it in the commit message.

wbclark (approved):

PR #295, opened 2020-11-16
Generate beaker setfiles on the fly

ekohl:
Draft PR to test out https://github.com/voxpupuli/puppet_metadata/pull/11. The goal is to avoid the duplication of setfiles. It modifies the cache building into a matrix generation step.


ekohl:
Will build a new one based on https://github.com/theforeman/foreman-installer-modulesync/pull/136.

PR #297, opened 2020-11-17
Refs #30436 - add import/export params

jeremylenz:
This is the "second PR" from https://github.com/theforeman/puppet-pulpcore/pull/147#pullrequestreview-529534099

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -338,6 +338,13 @@ class foreman_proxy_content (
   }
 
   if $pulpcore {
+    if $pulpcore_mirror {
+      $pulpcore_allowed_import_path    = ['/var/lib/pulp/sync_imports']
+      $pulpcore_allowed_export_path    = []
+    } else {
+      $pulpcore_allowed_import_path    = ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
+      $pulpcore_allowed_export_path    = ['/var/lib/pulp/exports']
+    }
     if $shared_with_foreman_vhost {
       include foreman::config::apache
       $servername = $foreman::config::apache::servername
@@ -373,6 +380,8 @@ class foreman_proxy_content (
     }
 
     class { 'pulpcore':
+      allowed_import_path       => $pulpcore_allowed_import_path,
+      allowed_export_path       => $pulpcore_allowed_export_path,
       apache_http_vhost         => $apache_http_vhost,
       apache_https_vhost        => $apache_https_vhost,
       apache_https_cert         => $apache_https_cert,
--- a/metadata.json
+++ b/metadata.json
@@ -38,7 +38,7 @@
     },
     {
       "name": "theforeman/pulpcore",
-      "version_requirement": ">= 2.1.0 < 3.0.0"
+      "version_requirement": ">= 2.2.0 < 3.0.0"
     }
   ],
   "requirements": [


wbclark (commented):
Thanks for kicking this off @jeremylenz !

What we also need here is to declare, set a default for, and document the new parameter we are introducing in this module. https://github.com/theforeman/puppet-foreman_proxy_content/commit/1d2eef1b71a5c1c5904f9447c05f7a3902cf5c1c is a nice and recent example of exactly what that looks like.

manifests/init.pp
```diff
@@ -369,6 +369,8 @@
     }
 
     class { 'pulpcore':
+      allowed_import_path       => $allowed_import_path,
+      allowed_export_path       => $allowed_export_path,
```
wbclark, line 373:
```suggestion
      allowed_import_path       => $pulpcore_allowed_import_path,
      allowed_export_path       => $pulpcore_allowed_export_path,
```

This keeps it clearer that we have 2 sets of parameters: `pulpcore::allowed_{import,export}_path` which is in the puppet-pulpcore module, and `foreman_proxy_content::pulpcore_allowed_{import,export}_path` here in puppet-FPC. The second pair only is what will be directly accessible by foreman-installer, so including pulpcore in the parameter name makes it easy to group that param with others that affect the pulpcore class.

jeremylenz: thanks for the help! :) Updated

manifests/init.pp
```diff
@@ -369,6 +369,8 @@
     }
 
     class { 'pulpcore':
+      allowed_import_path       => $pulpcore_allowed_import_path,
+      allowed_export_path       => $pulpcore_allowed_export_path,
```
ekohl, line 373: These variables are undefined.

ekohl (changes requested):

manifests/init.pp
wbclark: You can refer to https://github.com/theforeman/puppet-foreman_proxy_content/commit/1d2eef1b71a5c1c5904f9447c05f7a3902cf5c1c for an example of what this should look like

manifests/init.pp
```diff
@@ -108,6 +108,8 @@
 #
 # $pulpcore_postgresql_ssl_root_ca::    Path to the root certificate authority to validate the certificate supplied by the PostgreSQL database server.
 #
+# $pulpcore_allowed_import_path::       allowed import paths for pulpcore settings.py
+# $pulpcore_allowed_import_path::       allowed export paths for pulpcore settings.py
```
wbclark, line 112: If you wish to have some documentation about how these parameters are used in Katello / Content Proxy, you're allowed to be more specific here because this puppet module specifically exists for those use cases

manifests/params.pp
```diff
@@ -64,5 +64,7 @@
   $pulpcore_postgresql_ssl_cert    = '/etc/pki/katello/certs/pulpcore-database.crt'
   $pulpcore_postgresql_ssl_key     = '/etc/pki/katello/private/pulpcore-database.key'
   $pulpcore_postgresql_ssl_root_ca = '/etc/pki/tls/certs/ca-bundle.crt'
+  $pulpcore_allowed_import_path    = ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
+  $pulpcore_allowed_export_path    = ['/var/lib/pulp/exports']
```
wbclark, line 68:
```suggestion
  $pulpcore_allowed_import_path    = ['/var/lib/pulp/sync_imports']
  $pulpcore_allowed_export_path    = []
```

Let's set defaults here that most appropriate for the Content Proxy scenario (i.e. Capsule). This also makes sense to do, since the migration you have opened in https://github.com/theforeman/foreman-installer/pull/615 is for the Katello scenario only. So FPC scenario will have no migration and therefore just get the defaults for this module which we define here. Then for the Katello scenario we'll override the defaults with the installer migration and give the values that Katello needs instead.

I think this makes the most logical sense, and as an added benefit, it's consistent with what you are already doing at the installer level. :+1:

manifests/init.pp
wbclark: Also, you have `$pulpcore_allowed_import_path::` twice here. One should be `export` instead.

manifests/params.pp
ekohl: A different approach would be to hardcode it in `init.pp`, depending on whether it's a mirror or not. That would avoid the migrations in the installer itself. While it provides less flexibility for the user, it does make it easier to consume.

manifests/init.pp
jeremylenz: oops!  Updated

jeremylenz:
@wbclark updated

manifests/params.pp
```diff
@@ -64,5 +64,7 @@
   $pulpcore_postgresql_ssl_cert    = '/etc/pki/katello/certs/pulpcore-database.crt'
   $pulpcore_postgresql_ssl_key     = '/etc/pki/katello/private/pulpcore-database.key'
   $pulpcore_postgresql_ssl_root_ca = '/etc/pki/tls/certs/ca-bundle.crt'
+  $pulpcore_allowed_import_path    = ['/var/lib/pulp/sync_imports']
+  $pulpcore_allowed_export_path    = []
```
ehelms, line 68: I would set these to the defaults Katello expects, the values you have here -- https://github.com/theforeman/foreman-installer/pull/615/files#diff-f333bd70087fcf1988e26c39646174b3aac006d11e0412c84cbe081b0d573598R2-R3

jeremylenz: I had set them as per [this comment](https://github.com/theforeman/puppet-foreman_proxy_content/pull/297#discussion_r527992505) above.  should I change it back?

jeremylenz:
I _think_ the test failures are because https://github.com/theforeman/puppet-pulpcore/pull/147, which defines the `allowed_export_path` param, isn't merged yet.

manifests/params.pp
ehelms: I think https://github.com/theforeman/puppet-foreman_proxy_content/pull/297#discussion_r528764632 provides what I would envision. This module is focused heavily on deploying Pulp for Katello's use case. The difference is whether deploying in a mirror setup or not. 

Would a user ever need to configure these paths differently?

jeremylenz: I don't think so.  Updating

jeremylenz: @ehelms Changed back to your suggestion.  do I also need to make changes in `init.pp`?

manifests/params.pp
ehelms: What is being suggested here is more so to drop the parameters from being user defined and instead set them just prior to this based on the `pulpcore_mirror` condition, something like:

```
  if $pulpcore_mirror {
    $pulpcore_allowed_import_path    = ['/var/lib/pulp/sync_imports']
    $pulpcore_allowed_export_path    = []
  } else {
    $pulpcore_allowed_import_path    = ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
    $pulpcore_allowed_export_path    = ['/var/lib/pulp/exports']
  }
```

So if its a mirror, effectively set them to the defaults, otherwise if main Pulp set them to these needed import/export paths. And do this in `init.pp` before the pulpcore class is declared and not in params.pp.

manifests/init.pp
```diff
@@ -174,8 +174,6 @@
   Stdlib::Absolutepath $pulpcore_postgresql_ssl_cert = '/etc/pki/katello/certs/pulpcore-database.crt',
   Stdlib::Absolutepath $pulpcore_postgresql_ssl_key = '/etc/pki/katello/private/pulpcore-database.key',
   Stdlib::Absolutepath $pulpcore_postgresql_ssl_root_ca = '/etc/pki/tls/certs/ca-bundle.crt',
-  Array[Stdlib::Absolutepath] $pulpcore_allowed_import_path = $foreman_proxy_content::params::pulpcore_allowed_import_path,
-  Array[Stdlib::Absolutepath] $pulpcore_allowed_export_path = $foreman_proxy_content::params::pulpcore_allowed_export_path,
```
jeremylenz: Removed these lines because it wouldn't allow me to reassign these variables later.  Also it wouldn't allow me to _not_ assign them here and then assign them later.  But now, Github checks are failing with
```
manifests/init.pp:114:parameter_documentation:WARNING:No matching class parameter for documentation of foreman_proxy_content::pulpcore_allowed_export_path
```
because it's expecting the line I had to delete.  @ehelms what is the correct way forward?

manifests/init.pp
```diff
@@ -110,6 +110,8 @@
 #
 # $pulpcore_postgresql_ssl_root_ca::    Path to the root certificate authority to validate the certificate supplied by the PostgreSQL database server.
 #
+# $pulpcore_allowed_import_path::       allowed import paths for pulpcore settings.py. Used for Katello content view imports and repository sync with file protocol.
+# $pulpcore_allowed_export_path::       allowed export paths for pulpcore settings.py. Used for Katello content view exports.
```
ehelms, line 114: You can drop these now

jeremylenz: @ehelms thanks!  Updated and checks passing now :)

jeremylenz:
@ekohl any other concerns?

ekohl (changes requested):
It needs to update the dependency on puppet-pulpcore in metadata.json. I intend to do a release of it, but first https://github.com/theforeman/puppet-pulpcore/pull/150 should be merged. IMHO that should have been the implementation in the first place, but it was merged before I could correct it.

ekohl:
I opened https://github.com/theforeman/puppet-pulpcore/pull/151. Please update metadata.json to at least require 2.2.0.

jeremylenz:
@ekohl Thanks, updated

manifests/init.pp
```diff
@@ -337,6 +337,14 @@
     }
   }
 
+  if $pulpcore_mirror {
+    $pulpcore_allowed_import_path    = ['/var/lib/pulp/sync_imports']
+    $pulpcore_allowed_export_path    = []
+  } else {
+    $pulpcore_allowed_import_path    = ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports']
+    $pulpcore_allowed_export_path    = ['/var/lib/pulp/exports']
+  }
```
ekohl, line 346: Small nit: I would expect this within the `if $pulpcore` block since the variables are not used elsewhere.

ekohl (changes requested):
Please squash the commits.

ehelms (approved):

PR #299, opened 2020-11-23
Add enable Pulp 2 and Pulpcore parameters

ehelms:
@ekohl I wanted to see if this what you were thinking before I went further.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -20,11 +20,20 @@
 #
 # === Advanced parameters:
 #
+# $enable_pulpcore::                    Enables Pulpcore deployment
+#
+# $enable_pulpcore_mirror::             Whether this pulpcore deployment acts as a mirror.
+#                                       A pulpcore mirror is the pulpcore equivalent of a pulpnode.
+#
+# $enable_pulp2::                       Enables Pulp 2 deployment
+#
+# $enable_pulp2_mirror::                Enables Pulp 2 deployment as a mirror, known in Pulp 2 as a pulpnode
+#
 # $puppet::                             Enable puppet
 #
 # $pulp_admin_password::                Password for the Pulp admin user. It should be left blank so that a random password is generated
 #
-# $pulp_max_speed::                     The maximum download speed per second for a Pulp task, such as a sync. (e.g. "4 Kb" (Uses SI KB), 4MB, or 1GB" )
+# $pulp_max_speed::                     The maximum download speed per second for a Pulp task, such as a sync. (e.g. "4 Kb" (Uses SI KB), 4MB, or 1GB)
 #
 # $pulp_num_workers::                   Number of Pulp workers to use.
 #
@@ -116,6 +125,10 @@
 #                                       incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
 class foreman_proxy_content (
+  Boolean $enable_pulpcore = true,
+  Boolean $enable_pulpcore_mirror = false,
+  Boolean $enable_pulp2 = true,
+  Boolean $enable_pulp2_mirror = false,
   String[1] $parent_fqdn = $foreman_proxy_content::params::parent_fqdn,
   String $pulp_admin_password = $foreman_proxy_content::params::pulp_admin_password,
   Optional[String] $pulp_max_speed = $foreman_proxy_content::params::pulp_max_speed,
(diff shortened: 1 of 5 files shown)


ekohl:
This is indeed what I was thinking about

manifests/init.pp
```diff
@@ -116,6 +125,10 @@
 #                                       incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
 class foreman_proxy_content (
+  Boolean $enable_pulpcore = true,
+  Boolean $enable_pulpcore_mirror = false,
+  Boolean $enable_pulp2 = true,
+  Boolean $enable_pulp2_mirror = false,
```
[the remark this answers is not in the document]
ehelms: For me, that more about nomenclature. We want to stick with `pulpcore` based parameters going forward so having a `pulp` parameter that is for Pulp 2 and 3 makes it an oddball. I'd rather just have both and drop the Pulp 2 one when the time is right (which is rather close).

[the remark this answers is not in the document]
ekohl: Thoughts about `$content_is_mirror`?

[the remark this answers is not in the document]
ehelms: If it was internal, sure. But for a user, there is good parity in `enable_pulpcore` `enable_pulpcore_mirror`. I think there is an argument to be made for, if puppet-foreman_proxy_content is enabled, you get Pulpcore automatically as its implied. Which means we could reduce this to just the parameter `content_mirror` or `enable_content_mirror` ? But, given we already expose everything as `pulpcore` (https://github.com/theforeman/puppet-foreman_proxy_content/pull/299/files#diff-c6c3a824d185ff01e063011517a1b13a604dc3d7b63d1c67a30451c10df02bcbR100-R125) I am inclined to just stay with that nomenclature.

manifests/init.pp
```diff
@@ -20,11 +20,20 @@
 #
 # === Advanced parameters:
 #
+# $enable_pulpcore::                    Enables Pulpcore deployment
+#
+# $enable_pulpcore_mirror::             Whether this pulpcore deployment acts as a mirror.
```
ekohl, line 25: Is it pulpcore or Pulpcore? Probably good to be consistent in that.

manifests/init.pp
```diff
@@ -176,23 +189,17 @@
...
 
-  $pulp_master = $foreman_proxy::plugin::pulp::enabled
-  $pulp = $foreman_proxy::plugin::pulp::pulpnode_enabled
-  $pulpcore_mirror = $foreman_proxy::plugin::pulp::pulpcore_mirror
-  $pulpcore = $foreman_proxy::plugin::pulp::pulpcore_enabled
-
-  $enable_pulp2_rpm = $enable_yum and !($pulpcore and $proxy_pulp_yum_to_pulpcore)
-  $enable_pulp2_iso = $enable_file and !($pulpcore and $proxy_pulp_isos_to_pulpcore)
-  $enable_pulp2_deb = $enable_deb and !($pulpcore and $proxy_pulp_deb_to_pulpcore)
+  $enable_pulp2_rpm = $enable_yum and ($enable_pulp2 and !$proxy_pulp_yum_to_pulpcore)
```
ekohl, line 193: Nit: the braces are redundant now
```suggestion
  $enable_pulp2_rpm = $enable_yum and $enable_pulp2 and !$proxy_pulp_yum_to_pulpcore
```

manifests/init.pp
```diff
@@ -201,10 +208,17 @@
...
   ensure_packages('katello-debug')
 
-  if ($pulp_master or $pulp) and $facts['os']['release']['major'] != '7' {
+  if $enable_pulp2 and $facts['os']['release']['major'] != '7' {
     fail('Pulp 2 is only supported on CentOS 7')
   }
 
+  class { 'foreman_proxy::plugin::pulp':
+    enabled          => $enable_pulp2,
+    pulpnode_enabled => $enable_pulp2_mirror,
```
ekohl, line 217: This logic is incorrect. In Pulp 2 there was one plugin for pulp and one for pulpnode. Enabling both at the same time is illogical and probably confuses Katello.
```suggestion
    enabled          => $enable_pulp2 and !$enable_pulp2_mirror,
    pulpnode_enabled => $enable_pulp2 and $enable_pulp2_mirror,
```

manifests/init.pp
```diff
@@ -242,7 +258,7 @@
       }
       contain foreman_proxy_content::dispatch_router
 
-      if $pulp_master {
+      unless $enable_pulp2_mirror {
```
ekohl, line 261: Can you use `unless` if there is also an `else`? Maybe it was Ruby where that's not possible.

spec/classes/foreman_proxy_content_spec.rb
```diff
@@ -5,17 +5,13 @@
...
-          }
-          PUPPET
-        end
+      let(:params) do
+        {
+          enable_pulp2: false
+        }
+      end
 
+      context 'without parameters' do
```
ekohl, line 14: Technically not really correct anymore since you do set `params` by default.

spec/classes/foreman_proxy_content_spec.rb
```diff
@@ -24,18 +20,11 @@ class { 'foreman_proxy::plugin::pulp':
...
-        let(:pre_condition) do
-          <<-PUPPET
-          include foreman_proxy
-          class { 'foreman_proxy::plugin::pulp':
-            enabled          => false,
-            pulpnode_enabled => true,
+            qpid_router: false,
+            enable_pulp2: true,
+            enable_pulpcore: false,
+            enable_pulp2_mirror: true
```
ekohl, line 26: Mind sorting these?

manifests/init.pp
```diff
@@ -176,23 +189,17 @@
...
+  $enable_pulp2_rpm = $enable_yum and ($enable_pulp2 and !$proxy_pulp_yum_to_pulpcore)
+  $enable_pulp2_iso = $enable_file and ($enable_pulp2 and !$proxy_pulp_isos_to_pulpcore)
+  $enable_pulp2_deb = $enable_deb and ($enable_pulp2 and !$proxy_pulp_deb_to_pulpcore)
 
   $foreman_proxy_fqdn = $facts['networking']['fqdn']
   $foreman_url = $foreman_proxy::foreman_base_url
-  $reverse_proxy_real = ($pulp or $pulpcore_mirror) or $reverse_proxy
+  $reverse_proxy_real = ($enable_pulp2_mirror or $enable_pulpcore_mirror) or $reverse_proxy
 
   # TODO: doesn't allow deploying a Pulp non-mirror without Foreman
```
ekohl, line 201: I'm still thinking how we can achieve this but it's probably easier once we get rid of Pulp 2.

ekohl (changes requested):
Some implementation details inline, but overall I think this looks like it's heading in the right direction.

ehelms:
The more I think about this the more I am seeing these additional parameters as temporary and not needed. If we look at the current nightly situation for Katello we have the following:

 * Server: Should be Pulp 3 for content, Pulp 2 existing only to support katello-agent
 * Proxy: Pulp 3 only, with dispatch router present for katello-agent

We should be able to assume that if you are enabling `foreman_proxy_content` you are getting Pulp 3. The only option is whether or not this is a mirror installation. 

For Qpid, I think we could consider defaulting it to off on server and proxy and having users enable it going forward if they want that functionality. Especially as we consider the move is to drop Pulp 2 and have Katello own communication over the Qpid-katello-agent protocol.

ekohl:
Long term I think there are 3 deployment methods. In all of them, there should be Pulp 3.

1. Primary
.1. Deployed as part of the Foreman vhost
.2. Deployed standalone
2. Mirror
.1. Deployed standalone

Then there is the question of additional services. I think this is a summary of the status.

1.1. Pulp 2 can exist if desired for katello-agent. If so, should have qpid
1.2. Can't be properly supported until we drop Pulp 2 and Qpid?
2.1. Qpid dispatch router can be present if desired for katello-agent

Consider a scenario where server is EL7 but Smart Proxy is EL8. Let's say katello-agent is desired so the server has qpid. How does Katello know that qdrouterd is running on a Smart Proxy and can reach clients?

PR #300, opened 2020-11-23
Don't enable Pulp2 Deb with Pulpcore

wbclark:
This fixes an issue in https://github.com/theforeman/puppet-foreman_proxy_content/commit/1d2eef1b71a5c1c5904f9447c05f7a3902cf5c1c

CC: @jlsherrill , we missed this one


ekohl (changes requested):
Would you mind using the full commit ID instead of the short one? That way it's clickable in the Github UI. Otherwise :+1:

wbclark:
@ekohl thanks, it is updated

ekohl (approved):

PR #303, opened 2020-12-04
Set registry url to point at the Katello API endpoint

ehelms:
Requires: https://github.com/theforeman/puppet-pulpcore/pull/152

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -413,7 +413,11 @@ class foreman_proxy_content (
       }
     }
 
-    include pulpcore::plugin::container
+    unless $pulpcore_mirror {
+      class { 'pulpcore::plugin::container':
+        registry_base_url => $foreman::config::apache::_proxy_backend,
+      }
+    }
     class { 'pulpcore::plugin::file':
       use_pulp2_content_route => $proxy_pulp_isos_to_pulpcore,
     }


manifests/init.pp
```diff
@@ -413,7 +413,11 @@
       }
     }
 
-    include pulpcore::plugin::container
+    unless $pulpcore_mirror {
+      class { 'pulpcore::plugin::container':
+        registry_base_url => $foreman::config::apache::_proxy_backend,
```
ekohl, line 418: I think it would be cleaner to handle this in the `if $shared_with_foreman_vhost {` tree. Whether it's a mirror is irrelevant: what counts is how we configure Apache.

That is something I feel very strongly about, we need to get rid of that. In the future I want be able to deploy a non-mirror on a server where Foreman doesn't run. We should now be making sure that we don't introduce more blockers to this. Once we have that in place, we can create a separate scenario for this. Add shared storage for `/var/lib/pulp`, external PostgreSQL, Redis and a load balancer and you have a horizontally scalable deployment. I'm certain a few downstream customers would appreciate this.

Similarly, you should be able to do the same thing for mirror Pulp instances to horizontally scale them out if needed.

ekohl (changes requested):

PR #307, opened 2020-12-14
Fixes #31509 - Allow changing node FQDN

ekohl:
The Katello certificates default to using the FQDN. This is done in the certs class via the node_fqdn parameter. The specific classes have a parameter named hostname which defaults to $certs::node_fqdn.

Prior to this patch, those hostnames were hardcoded to the FQDN fact which made it impossible to decouple the system hostname from the service name. With this change, it is possible to do so.

The Puppet certificate is matched to the Foreman Proxy since these in fact must have the same name. Otherwise Foreman authentication will fail. That is because the Puppet scripts "authenticate" via the Smart Proxy registration by using the same Common Name on the certificate.


ehelms (approved):
let's take it for a spin

PR #309, opened 2020-12-16
Drop parent_fqdn in favor of deriving from foreman_url

ehelms:
I wanted to continue working toward the number of things a user has to specify to get things working and in this case the number of things a user has to specify multiple times, via different parameters. If we take https://github.com/theforeman/forklift/blob/master/pipelines/install/04-install_proxy.yaml#L11 as an example, the user effectively specifies the Foreman server hostname 4 times in the most basic of commands to install a content proxy.

### DIFF
--- a/lib/puppet/functions/foreman_proxy_content/host_from_url.rb
+++ b/lib/puppet/functions/foreman_proxy_content/host_from_url.rb
@@ -1,0 +1,19 @@
+require 'uri'
+
+# @summary Parses a URL and returns the host
+#
+# @example Calling the function
+#   $host = foreman_proxy_content::host_from_url('https://test.example.com')
+Puppet::Functions.create_function(:'foreman_proxy_content::host_from_url') do
+  # @param url URL to parse
+  # @return The host found within the URL
+  dispatch :host do
+    param 'Stdlib::HTTPUrl', :url
+    return_type 'Stdlib::Host'
+  end
+
+  def host(url)
+    uri = URI(url)
+    uri.host
+  end
+end
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -4,8 +4,6 @@
 #
 # === Parameters:
 #
-# $parent_fqdn::                        FQDN of the parent node.
-#
 # $enable_ostree::                      Enable ostree content plugin, this requires an ostree install
 #
 # $enable_yum::                         Enable rpm content plugin, including syncing of yum content
@@ -114,7 +112,6 @@
 #                                       incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
 class foreman_proxy_content (
-  String[1] $parent_fqdn = $foreman_proxy_content::params::parent_fqdn,
   String $pulp_admin_password = $foreman_proxy_content::params::pulp_admin_password,
   Optional[String] $pulp_max_speed = undef,
   Optional[Integer[1]] $pulp_num_workers = undef,
@@ -185,6 +182,7 @@ class foreman_proxy_content (
   $enable_pulp2_deb = $enable_deb and !($pulpcore and $proxy_pulp_deb_to_pulpcore)
 
   $foreman_url = $foreman_proxy::foreman_base_url
+  $foreman_host = foreman_proxy_content::host_from_url($foreman_url)
   $reverse_proxy_real = ($pulp or $pulpcore_mirror) or $reverse_proxy
 
   # TODO: doesn't allow deploying a Pulp non-mirror without Foreman
@@ -242,7 +240,7 @@ class foreman_proxy_content (
         contain foreman_proxy_content::dispatch_router::hub
       } else {
         class { 'foreman_proxy_content::dispatch_router::connector':
-          host => $parent_fqdn,
+          host => $foreman_host,
           port => $qpid_router_hub_port,
         }
         contain foreman_proxy_content::dispatch_router::connector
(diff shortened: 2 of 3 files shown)


manifests/init.pp
```diff
@@ -185,6 +182,7 @@
   $enable_pulp2_deb = $enable_deb and !($pulpcore and $proxy_pulp_deb_to_pulpcore)
 
   $foreman_url = $foreman_proxy::foreman_base_url
+  $foreman_hostname = regsubst($foreman_url, 'https://', '')
```
ekohl, line 185: The Foreman URL can really be a URL and include a path. I think the safest would be to add a function to extract the hostname from a URL. https://github.com/voxpupuli/puppet-extlib/blob/master/lib/puppet/functions/extlib/dump_args.rb is probably the best example of a minimal function. Then use Ruby's URL parsing to do so.

ekohl (changes requested):

manifests/init.pp
ehelms: Hah. I almost went that route. Then said... let me try the 'dumb' route and save myself time just in case :)

evgeni: I wonder if that's actually true for Katello deployments, or if we hardcore paths like /rhsm on clients so that you can't deploy it in a "folder".

ekohl: Still, writing your own parsing is a likely a source of problems. For example, even if we only support `/`, the current code doesn't deal with `https://foreman.example.com/` which is a valid input.

ehelms:
Now with a function for deriving the hostname by parsing the URL. Admittedly, I couldn't come up with a better name than `hostname` for the function. Open to suggestions if others don't like it.

lib/puppet/functions/foreman_proxy_content/hostname.rb
```diff
@@ -0,0 +1,20 @@
...
+# @summary Parses a URL and returns the hostname
+#
+# @example Calling the function
+#   $hostname = hostname('https://test.example.com')
+Puppet::Functions.create_function(:'foreman_proxy_content::hostname') do
+  # @param url URL to parse
+  # @return The hostname found within the URL
+  dispatch :hostname do
+    param 'Stdlib::HTTPUrl', :url
+    return_type 'Any'
```
ekohl, line 12:
```suggestion
    return_type 'Stdlib::Host'
```

spec/functions/foreman_proxy_content/hostname_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+describe 'foreman_proxy_content::hostname' do
+
+	it { expect(subject).not_to eq(nil) }
+	it { is_expected.not_to eq(nil) }
+	it { is_expected.to run.with_params('').and_raise_error(ArgumentError) }
+	it { is_expected.to run.with_params('just a string').and_raise_error(ArgumentError) }
+
+	context 'with basic URL' do
+		let(:url) { 'https://test.example.com' }
```
ekohl, line 11: Looks like you're using tabs while everywhere else we use spaces.

ekohl (commented):
`url2hostname` perhaps?

lib/puppet/functions/foreman_proxy_content/hostname_from_url.rb
```diff
@@ -0,0 +1,19 @@
+require 'uri'
+
+# @summary Parses a URL and returns the hostname
+#
+# @example Calling the function
+#   $hostname = hostname('https://test.example.com')
```
ekohl, line 6: Functions need to be called with their namespace
```suggestion
#   $hostname = foreman_proxy_content::hostname_from_url('https://test.example.com')
```

lib/puppet/functions/foreman_proxy_content/hostname_from_url.rb
```diff
@@ -0,0 +1,19 @@
...
+  # @param url URL to parse
+  # @return The hostname found within the URL
+  dispatch :hostname do
+    param 'Stdlib::HTTPUrl', :url
+    return_type 'Stdlib::Host'
+  end
+
+  def hostname(url)
+    uri = URI(url)
+    uri.hostname
```
ekohl, line 17: Interesting that https://ruby-doc.org/stdlib-2.7.2/libdoc/uri/rdoc/URI.html makes no mention of `hostname`. There is only host.

ekohl (changes requested):
Looking at the URI docs and RFC I think we should use `host`. `hostname` implies it's a name, but technically an IP is also valid. Yes, I know that `ip:` is rare in `subjectAltName`, but let's follow the standard. We can also handle that fine in the other code.

lib/puppet/functions/foreman_proxy_content/host_from_url.rb
```diff
@@ -0,0 +1,19 @@
+require 'uri'
+
+# @summary Parses a URL and returns the host
+#
+# @example Calling the function
+#   $host = host_from_url('https://test.example.com')
```
ekohl, line 6: This still needs the namespace
```suggestion
#   $host = foreman_proxy_content::host_from_url('https://test.example.com')
```

manifests/init.pp
```diff
@@ -185,6 +182,7 @@
   $enable_pulp2_deb = $enable_deb and !($pulpcore and $proxy_pulp_deb_to_pulpcore)
 
   $foreman_url = $foreman_proxy::foreman_base_url
+  $foreman_hostname = foreman_proxy_content::host_from_url($foreman_url)
```
ekohl, line 185: Let's make this host too
```suggestion
  $foreman_host = foreman_proxy_content::host_from_url($foreman_url)
```

ekohl (commented):
2 small nits, otherwise :+1:

ekohl (approved):

ekohl:
This is backwards incompatible and we will need to handle this in forklift where we pass it in our CI. Do you want to wait for that PR to be ready before we merge it?

PR #315, opened 2021-01-19
Fixes #31662: Set enable_http to ensure pub dir is deployed to Pulp v…

ehelms:
…host

This should be cherry picked to 3.18. This affects only content proxies deployed in Katello 3.18 (or current nightly, all that changes once we go Pulp 3 only) that have `enable_puppet` set to false. The reason is this chunk of code: https://github.com/theforeman/puppet-pulp/blob/master/manifests/apache.pp#L23

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -288,6 +288,7 @@ class foreman_proxy_content (
       enable_deb             => $enable_pulp2_deb,
       enable_puppet          => $enable_puppet,
       enable_docker          => $enable_docker,
+      enable_http            => true,
       default_password       => $pulp_admin_password,
       messaging_transport    => 'qpid',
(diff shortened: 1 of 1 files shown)


evgeni:
By "pub dir" you mean `/var/www/pub/` which hosts all those pulp bits, *NOT* `forman.example.com/pub` which only hosts ca cert etc, right? If so, then ACK

ehelms:
Nope, I mean `foreman.example.com/pub` where bootstrap RPM and such live. The user issue thread if it helps to understand -- https://community.theforeman.org/t/content-proxy-not-listing-to-port-80-anymore/21996

evgeni:
Ah, the issue IS actually about the cert...

evgeni (approved):

PR #316, opened 2021-01-28
Ensure foreman_proxy::plugin::pulp does not enable Pulp 2 on EL8

ehelms:


manifests/init.pp
```diff
@@ -273,7 +273,7 @@
   include pulpcore::plugin::certguard # Required to be present by Katello when syncing a content proxy
 
   class { 'foreman_proxy::plugin::pulp':
-    enabled              => !$pulpcore_mirror,
+    enabled              => !$pulpcore_mirror and $facts['os']['release']['major'] == '7',
```
wbclark, line 276: I think this would create an issue for non-RH operating systems. What should the behavior be here for those users?

ekohl (approved):

manifests/init.pp
ekohl: This module doesn't support non-RH OSes so it's not really a factor.

ehelms: This isn't supported anywhere but EL operating systems so we are good.

PR #318, opened 2021-01-29
Drop unused Pulp dependency and pulp_nodes template

ekohl:
018a88a58c19672d24899fddd529b55dc7ddcc3e removed the last use of the module. It also drops the development dependencies introduced only for pulp.

f68ee549f0b297e940a81442843ddadc988c6347 removed the last use of the template.


ehelms (approved):

ehelms:
This seemed to fix the EL7 proxy still having a pulp-deb-plugins being installed.

PR #319, opened 2021-01-29
Fixes #31642 - Add container gateway support

ianballou:
Supports the Container Gateway smart proxy plugin by adding the proper Apache proxy configuration.

Fragment example that appears at the end of the 443 vhost:
```
  ## Container Gateway
  <Location "/pulpcore_registry/v2/">
    Require expr %{SSL_CLIENT_S_DN_CN} == "centos7-proxy-devel-3.cannolo.example.com"
    RequestHeader set SSL_CLIENT_S_DN "admin"
  </Location>
  ProxyPass /v1/ https://127.0.0.1:9090/container_gateway/v1/
  ProxyPassReverse /v1/ https://127.0.0.1:9090/container_gateway/v1/
  ProxyPass /v2/ https://127.0.0.1:9090/container_gateway/v2/
  ProxyPassReverse /v2/ https://127.0.0.1:9090/container_gateway/v2/
  SSLProxyCheckPeerCN off
  SSLProxyCheckPeerName off
```

I'm pretty new to the foreman-installer so let me know any places where I'm breaking standards.

TODO:
- [x] Add tests

### DIFF
--- a/manifests/container.pp
+++ b/manifests/container.pp
@@ -1,0 +1,46 @@
+# @summary Pulp Container plugin
+# @param location_prefix
+#   In the Apache configuration a location with this prefix is exposed. The
+#   version will be appended.
+# @param registry_v1_path
+#   The path beneath the location prefix to forward. This is also appended to
+#   the content base url.
+# @param registry_v2_path
+#   The path beneath the location prefix to forward. This is also appended to
+#   the content base url.
+class foreman_proxy_content::container (
+  String $location_prefix = '/pulpcore_registry',
+  String $registry_v1_path = '/v1/',
+  String $registry_v2_path = '/v2/',
+) {
+  include certs::foreman_proxy
+
+  $context = {
+    'directories'  => [
+      {
+        'provider'        => 'location',
+        'path'            => "${location_prefix}${registry_v2_path}",
+        'request_headers' => ["set SSL_CLIENT_S_DN \"admin\""],
+        'requires'        => ["expr %{SSL_CLIENT_S_DN_CN} == \"${certs::foreman_proxy::hostname}\""]
+      },
+    ],
+    'proxy_pass' => [
+      {
+        'path' => $registry_v1_path,
+        'url'  => "${foreman_proxy::real_registered_proxy_url}/container_gateway${registry_v1_path}",
+      },
+      {
+        'path' => $registry_v2_path,
+        'url'  => "${foreman_proxy::real_registered_proxy_url}/container_gateway${registry_v2_path}",
+      },
+    ],
+  }
+
+  $https_content = epp('foreman_proxy_content/container-gateway-fragment.epp', $context)
+
+  apache::vhost::fragment { 'pulp-https-container':
+    vhost    => 'pulpcore-https',
+    priority => '10',
+    content  => $https_content,
+  }
+}
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -254,6 +254,12 @@ class foreman_proxy_content (
 
   if $enable_docker {
     include pulpcore::plugin::container
+    unless $shared_with_foreman_vhost {
+      include foreman_proxy_content::container
+      class { 'foreman_proxy::plugin::container_gateway':
+        pulp_endpoint => "https://${servername}",
+      }
+    }
   }
   if $enable_file {
     class { 'pulpcore::plugin::file':
--- a/metadata.json
+++ b/metadata.json
@@ -22,7 +22,7 @@
     },
     {
       "name": "theforeman/foreman_proxy",
-      "version_requirement": ">= 13.0.0 < 18.0.0"
+      "version_requirement": ">= 17.1.0 < 18.0.0"
     },
     {
       "name": "katello/qpid",
--- a/spec/classes/foreman_proxy_content__container_spec.rb
+++ b/spec/classes/foreman_proxy_content__container_spec.rb
@@ -1,0 +1,51 @@
+require 'spec_helper'
+
+describe 'foreman_proxy_content::container' do
+  on_supported_os.each do |os, facts|
+    context "on #{os}" do
+      let(:facts) { facts }
+      let(:pre_condition) do
+        <<-PUPPET
+        include foreman_proxy
+        PUPPET
+      end
+
+      describe 'with inherited parameters' do
+
+        it { is_expected.to compile.with_all_deps }
+        it do
+          is_expected.to contain_apache__vhost__fragment('pulp-https-container')
+            .with_vhost('pulpcore-https')
+            .with_priority('10')
+            .with_content(%r{^\s+<Location "/pulpcore_registry/v2/">$})
+            .with_content(%r{^\s+Require expr %\{SSL_CLIENT_S_DN_CN\} == "foo.example.com"$})
+            .with_content(%r{^\s+RequestHeader set SSL_CLIENT_S_DN "admin"$})
+            .with_content(%r{^\s+</Location>$})
+            .with_content(%r{^\s+ProxyPass /v1/ https://foo\.example\.com:8443/container_gateway/v1/$})
+            .with_content(%r{^\s+ProxyPass /v2/ https://foo\.example\.com:8443/container_gateway/v2/$})
+            .with_content(%r{^\s+ProxyPassReverse /v1/ https://foo\.example\.com:8443/container_gateway/v1/$})
+            .with_content(%r{^\s+ProxyPassReverse /v2/ https://foo\.example\.com:8443/container_gateway/v2/$})
+        end
+      end
+
+      describe 'with explicit parameters' do
+        let(:params) { { location_prefix: '/other_pulpcore_registry', registry_v1_path: '/vr1/', registry_v2_path: '/vr2/' } }
+
+        it { is_expected.to compile.with_all_deps }
+        it do
+          is_expected.to contain_apache__vhost__fragment('pulp-https-container')
+            .with_vhost('pulpcore-https')
+            .with_priority('10')
+            .with_content(%r{^\s+<Location "/other_pulpcore_registry/vr2/">$})
+            .with_content(%r{^\s+Require expr %\{SSL_CLIENT_S_DN_CN\} == "foo.example.com"$})
+            .with_content(%r{^\s+RequestHeader set SSL_CLIENT_S_DN "admin"$})
+            .with_content(%r{^\s+</Location>$})
+            .with_content(%r{^\s+ProxyPass /vr1/ https://foo\.example\.com:8443/container_gateway/vr1/$})
+            .with_content(%r{^\s+ProxyPass /vr2/ https://foo\.example\.com:8443/container_gateway/vr2/$})
+            .with_content(%r{^\s+ProxyPassReverse /vr1/ https://foo\.example\.com:8443/container_gateway/vr1/$})
+            .with_content(%r{^\s+ProxyPassReverse /vr2/ https://foo\.example\.com:8443/container_gateway/vr2/$})
+        end
+      end
+    end
+  end
+end
--- a/templates/container-gateway-fragment.epp
+++ b/templates/container-gateway-fragment.epp
@@ -1,0 +1,37 @@
+<%- |
+  Array[Struct[{
+    provider => Enum['directory', 'directorymatch', 'location', 'locationmatch', 'files', 'filesmatch', 'proxy', 'proxymatch'],
+    path => String[1],
+    Optional[request_headers] => Array[String[1]],
+    Optional[requires] => Array[String[1]],
+  }]] $directories,
+  Array[Hash[String, String]] $proxy_pass = [],
+  Array[String] $misc_options = [],
+| -%>
+
+  ## Container Gateway
+<%- $directories.each |$directory| { -%>
+  <%-
+    if $directory['provider'] =~ /^(.*)match$/ {
+      $provider = $1.capitalize + 'Match'
+    } else {
+      $provider = $directory['provider'].capitalize
+    }
+  -%>
+  <<%= $provider %> "<%= $directory['path'] %>">
+    <%- if $directory['requires'] { -%>
+    <%-   $directory['requires'].each |$requires_statement| { -%>
+    Require <%= $requires_statement %>
+    <%-   } -%>
+    <%- } -%>
+    <%- if $directory['request_headers'] { -%>
+    <%-   $directory['request_headers'].each |$request_statement| { -%>
+    RequestHeader <%= $request_statement %>
+    <%-   } -%>
+    <%- } -%>
+  </<%= $provider %>>
+<% } -%>
+<% $proxy_pass.each |$proxy| { -%>
+  ProxyPass <%= $proxy['path'] %> <%= $proxy['url'] %>
+  ProxyPassReverse <%= $proxy['path'] %> <%= $proxy['url'] %>
+<% } -%>


ianballou:
Related PR: https://github.com/theforeman/puppet-foreman_proxy/pull/643

manifests/init.pp
```diff
@@ -254,6 +254,7 @@
 
   if $enable_docker {
     include pulpcore::plugin::container
+    include foreman_proxy_content::container
```
ekohl, line 257: We know the exact pulp URL so it would be better to include the foreman_proxy::plugin::
```puppet
    class { 'foreman_proxy::plugin::container_gateway':
      pulp_url => "https://${servername}",
    }
```

This means that if Pulp runs on a URL that's different from the system hostname, it'll still function correctly. It makes it more predictable: there's an answer (IIRC set via `--certs-hostname` and stored in answers). If you change the hostname (like with `hostnamectl`) and rerun the installer, it'll still continue to serve with the old hostname. Only until you update the certificates will you change everything end to end.

If we follow this path, we can slowly make `katello-change-hostname` less needed.

metadata.json
```diff
@@ -22,7 +22,7 @@
     },
     {
       "name": "theforeman/foreman_proxy",
-      "version_requirement": ">= 13.0.0 < 17.0.0"
+      "version_requirement": ">= 16.0.0 < 17.0.0"
```
ekohl, line 25: This is good, except that the addition of your plugin will probably be 16.1.0.

manifests/container.pp
```diff
@@ -0,0 +1,49 @@
...
+  String $registry_v2_path = '/v2/',
+) {
+
+  $context = {
+    'directories'  => [
+      {
+        'provider'        => 'location',
+        'path'            => "${location_prefix}${$registry_v2_path}",
+        'request_headers' => ["set SSL_CLIENT_S_DN \"admin\""],
+        'requires'        => ["expr %{SSL_CLIENT_S_DN_CN} == \"${facts['networking']['fqdn']}\""]
```
ekohl, line 23: This shouldn't be the FQDN but rather the hostname on the client cert right? So I'd `include certs::foreman_proxy` and then use `$certs::foreman_proxy::hostname`. Not sure if the exact variables are correct.

The short summary is: use variables from classes that actually ensure the resources.

manifests/container.pp
```diff
@@ -0,0 +1,49 @@
...
+      },
+    ],
+    'proxy_pass' => [
+      {
+        'path' => $registry_v1_path,
+        'url'  => "https://127.0.0.1:9090/container_gateway${$registry_v1_path}",
```
ekohl, line 29: Please reuse variables here
```suggestion
        'url'  => "${foreman_proxy::real_registered_proxy_url}/container_gateway${registry_v1_path}",
```

The same below. I think this also means you actually can check the peer certificates but you may need to provide the CA to verify the connection. I think that's done via `SSLProxyMachineCertificateFile`. I think the value is `$certs::foreman_proxy::default_ca`.

ekohl (changes requested):

manifests/init.pp
ianballou: After adding in the class like you mentioned above and enabling the plugin, I'm getting a duplicate declaration error. Does this class declaration somehow conflict with what's in the answers file for foreman-proxy-content?

manifests/container.pp
ianballou: Hm `$certs::foreman_proxy::default_ca` definitely exists, but it's printing out as `Ca[katello-default-ca]`.

Fixing the config so that it's 
```
SSLProxyMachineCertificateFile /etc/pki/katello/certs/katello-default-ca.crt
```
crashes httpd though, so something is missing here anyway.

manifests/init.pp
ekohl: Yes. I would suggest you don't include it in the answers file. We've done the same thing with the pulp plugin itself. This module is responsible for connecting them.

More background information: in Puppet there's a [roles and profiles method](https://puppet.com/docs/pe/2019.8/osp/the_roles_and_profiles_method.html). While it doesn't 100% map to our installer, it helps you understand a lot.

First of all, there are the component modules. They provide a base layer of functionality. Think apache, postgresql, redis. The end user of the installer doesn't directly see them but they're certainly there.

A layer up there are profile modules. They combine component modules in higher level functionality. This content module is that: they take a bunch of component modules and provide complete functionality. Another example is puppet-katello. We can also argue that puppet-foreman_proxy has some of those elements since it can manage a DHCP and DNS server as well.

In our installer we don't really have role modules. You could argue that an installer scenario is a role because it determines what can be enabled.

With that background in mind, let's apply it to the current case. We can expose all component modules to the user, but I'm arguing that instead we should bundle up the entire functionality in this module and only expose this module. We can still keep `foreman_proxy::plugin::container_gateway` and include it here, but the end user doesn't have to know about it. We can connect answers that we know that are related (`$pulp_url`) which also means they can't go out of sync.

Another way to look at it is whether it makes sense for a user to enable one part, but not the other. The container gateway plugin doesn't make a whole lot of sense without the other parts so exposing it via the answers file is probably not the right choice.

manifests/container.pp
ekohl: I'll need to take a more detailed look.

manifests/init.pp
ianballou: > You could argue that an installer scenario is a role because it determines what can be enabled.

After reading about the differences between roles and profiles I think I agree with this too...

> Another way to look at it is whether it makes sense for a user to enable one part, but not the other. The container gateway plugin doesn't make a whole lot of sense without the other parts so exposing it via the answers file is probably not the right choice.

Yeah I see what you mean about keeping the container gateway more closely connected to the parameters that are already configured.  It seemed a little off to me to reconfigure what we already know. I'll update it.

spec/classes/foreman_proxy_content__container_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+          include foreman_proxy_content
+          PUPPET
+        end
+
+        it { is_expected.to compile.with_all_deps }
+        it do
+          is_expected.to contain_apache__vhost__fragment('pulp-https-container')
+            .with_vhost('pulpcore-https')
+            .with_priority('10')
+          # TODO: test http_content
```
ianballou, line 25: Trying to figure out how I can test the `vhost` itself. Perhaps there's a way to validate `$https_content` ?

ianballou: Answered this myself: `.with_content(...)`

spec/classes/foreman_proxy_content__container_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+            .with_content(%r{^\s+ProxyPassReverse /v2/ https://foo\.example\.com:8443/container_gateway/v2/$})
+            .with_content(%r{^\s+SSLProxyCheckPeerCN off$})
+            .with_content(%r{^\s+SSLProxyCheckPeerName off$})
+        end
+      end
+
+      describe 'with explicit parameters' do
+        let(:params) { { location_prefix: '/different_pulpcore_registry', registry_v1_path: '/vr1', registry_v2_path: '/vr2' } }
+
+        it { is_expected.to compile.with_all_deps }
```
ianballou, line 40: Seeing this error locally for some reason:
```
Failure/Error: it { is_expected.to compile.with_all_deps }
       error during compilation: Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Class[Foreman_proxy_content::Container] is already declared; cannot redeclare (line: 8) (line: 8, column: 1) on node cannolo.usersys.redhat.com
     # ./spec/classes/foreman_proxy_content__container_spec.rb:40:in `block (5 levels) in <top (required)>'
```

Not sure yet where this "Duplicate declaration" is happening.

ianballou:
```
  error during compilation: Evaluation Error: Error while evaluating a Resource Statement, Could not find declared class foreman_proxy::plugin::container_gateway (file: /home/runner/work/puppet-foreman_proxy_content/puppet-foreman_proxy_content/spec/fixtures/modules/foreman_proxy_content/manifests/init.pp, line: 258, column: 5)
```

I'm assuming the tests aren't going to pass until the foreman-proxy PR is merged.

ekohl:
And that's a correct assumption.

manifests/container.pp
ianballou: Removing the `'SSLProxyCheckPeerCN off',  'SSLProxyCheckPeerName off',` bits works actually without any additions, guess I didn't test that before. Going to remove it and we should be all good there.

ianballou:
@ekohl I think I've addressed all the issues.

manifests/container.pp
ekohl: That's because `${foreman_proxy::real_registered_proxy_url}` contains the hostname that's on the certificate. Then the peer check actually passes. `127.0.0.1` is very unlikely to ever be on the certificate.

manifests/container.pp
```diff
@@ -0,0 +1,46 @@
...
+#   In the Apache configuration a location with this prefix is exposed. The
+#   version will be appended.
+# @param registry_v1_path
+#   The path beneath the location prefix to forward. This is also appended to
+#   the content base url.
+# @param registry_v2_path
+#   The path beneath the location prefix to forward. This is also appended to
+#   the content base url.
+class foreman_proxy_content::container (
+  String $location_prefix = '/pulpcore_registry',
```
ekohl, line 12: This path overlaps with https://github.com/theforeman/puppet-pulpcore/blob/e3e27dd380823ac696019ccb0f6cc64aefba99d5/manifests/plugin/container.pp#L9. It feels at least confusing but probably unreliable to have two reverse proxy configs for the same path. I think it's better to make `pulpcore::plugin::container` configurable so it accepts a different authentication method.

For what it's worth, I thought about doing the exact same thing for **all** pulpcore authentication to remove the need for a specific client certificate from the Katello side. Looks like I haven't filed an issue for it, but it's roughly the same as https://projects.theforeman.org/issues/30368. Foreman has a client certificate to connect to Smart Proxies. We can use the same client certificates for Pulp 3.

@jlsherrill @ehelms I think we've talked about this but haven't scheduled anything. It would greatly simplify a lot of things that we need to do regarding certificates.

ekohl (changes requested):

manifests/container.pp
ianballou: I suppose it would be better for all the directives to be under the same `<Location "/pulpcore_registry/v2/">`, but I believe the two `Location`s with the same path will just be merged (http://httpd.apache.org/docs/current/sections.html#merging).  I tried reversing the order and it doesn't seem to matter either.



> For what it's worth, I thought about doing the exact same thing for **all** pulpcore authentication to remove the need for a specific client certificate from the Katello side. Looks like I haven't filed an issue for it, but it's roughly the same as https://projects.theforeman.org/issues/30368. Foreman has a client certificate to connect to Smart Proxies. We can use the same client certificates for Pulp 3.
> 
> @jlsherrill @ehelms I think we've talked about this but haven't scheduled anything. It would greatly simplify a lot of things that we need to do regarding certificates.

Perhaps a goal for 4.1? Just want to make sure we're not late for 4.0.

metadata.json
```diff
@@ -22,7 +22,7 @@
     },
     {
       "name": "theforeman/foreman_proxy",
-      "version_requirement": ">= 13.0.0 < 18.0.0"
+      "version_requirement": ">= 16.1.0 < 17.0.0"
```
ekohl, line 25: Release PR @ https://github.com/theforeman/puppet-foreman_proxy/pull/644

ianballou: I've updated the floor to 17.1.0 and the ceiling back to 18

ekohl:
I started to take a stab at writing acceptance tests (or rather, update my WIP ones). From there I got the following virtual host:
```
# ************************************
# Vhost template in module puppetlabs-apache
# Managed by Puppet
# ************************************
#
<VirtualHost *:443>
  ServerName centos7-64.example.com

  ## Vhost docroot
  DocumentRoot "/var/lib/pulp/pulpcore_static"

  ## Directories, there should at least be a declaration for /var/lib/pulp/pulpcore_static

  <Directory "/var/lib/pulp/pulpcore_static">
    Options -Indexes -FollowSymLinks
    AllowOverride None
    Require all granted
  </Directory>

  <Location "/pulp/content">
    Require all granted
      ## Request Header rules
            RequestHeader unset X-CLIENT-CERT
            RequestHeader set X-CLIENT-CERT "%{SSL_CLIENT_CERT}s" env=SSL_CLIENT_CERT
    ProxyPass unix:///run/pulpcore-content.sock|http://pulpcore-content/pulp/content disablereuse=on timeout=600
    ProxyPassReverse unix:///run/pulpcore-content.sock|http://pulpcore-content/pulp/content
        </Location>

  <Location "/pulp/api/v3">
    Require all granted
      ## Request Header rules
            RequestHeader unset REMOTE_USER
            RequestHeader set REMOTE_USER "%{SSL_CLIENT_S_DN_CN}s" env=SSL_CLIENT_S_DN_CN
    ProxyPass unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp/api/v3 timeout=600
    ProxyPassReverse unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp/api/v3
        </Location>

  ## Logging
  ErrorLog "/var/log/httpd/pulpcore-https_error_ssl.log"
  ServerSignature Off
  CustomLog "/var/log/httpd/pulpcore-https_access_ssl.log" combined
  ProxyPass /katello/api/v2/repositories/ https://centos7-64.example.com/katello/api/v2/repositories/
  <Location /katello/api/v2/repositories/>
    ProxyPassReverse https://centos7-64.example.com/
  </Location>
  SSLProxyEngine On

  <Location "/pulpcore_registry/v2/">
    RequestHeader unset REMOTE_USER
    RequestHeader set REMOTE_USER "%{SSL_CLIENT_S_DN_CN}s" env=SSL_CLIENT_S_DN_CN
    ProxyPass unix:///run/pulpcore-api.sock|http://pulpcore-api/v2/
    ProxyPassReverse unix:///run/pulpcore-api.sock|http://pulpcore-api/v2/
  </Location>

  ProxyPass /pulp/container/ unix:///run/pulpcore-content.sock|http://pulpcore-content/pulp/container/
  ProxyPassReverse /pulp/container/ unix:///run/pulpcore-content.sock|http://pulpcore-content/pulp/container/

  <Location "/pulp/deb">
    RequestHeader unset X-CLIENT-CERT
    RequestHeader set X-CLIENT-CERT "%{SSL_CLIENT_CERT}s" env=SSL_CLIENT_CERT
    ProxyPass unix:///run/pulpcore-content.sock|http://pulpcore-content/pulp/content disablereuse=on timeout=600
    ProxyPassReverse unix:///run/pulpcore-content.sock|http://pulpcore-content/pulp/content
  </Location>

  <Location "/pulp/isos">
    RequestHeader unset X-CLIENT-CERT
    RequestHeader set X-CLIENT-CERT "%{SSL_CLIENT_CERT}s" env=SSL_CLIENT_CERT
    ProxyPass unix:///run/pulpcore-content.sock|http://pulpcore-content/pulp/content disablereuse=on timeout=600
    ProxyPassReverse unix:///run/pulpcore-content.sock|http://pulpcore-content/pulp/content
  </Location>

  <Location "/pulp/repos">
    RequestHeader unset X-CLIENT-CERT
    RequestHeader set X-CLIENT-CERT "%{SSL_CLIENT_CERT}s" env=SSL_CLIENT_CERT
    ProxyPass unix:///run/pulpcore-content.sock|http://pulpcore-content/pulp/content disablereuse=on timeout=600
    ProxyPassReverse unix:///run/pulpcore-content.sock|http://pulpcore-content/pulp/content
  </Location>
Alias /pub /var/www/html/pub

<Location /pub>
  <IfModule mod_passenger.c>
    PassengerEnabled off
  </IfModule>
  Options +FollowSymLinks +Indexes
  Require all granted
</Location>

  ## Proxy rules
  ProxyRequests Off
  ProxyPreserveHost Off
  ProxyPass /pulp/assets/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp/assets/
  ProxyPassReverse /pulp/assets/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp/assets/

  ## SSL directives
  SSLEngine on
  SSLCertificateFile      "/etc/pki/katello/certs/katello-apache.crt"
  SSLCertificateKeyFile   "/etc/pki/katello/private/katello-apache.key"
  SSLCertificateChainFile "/etc/pki/katello/certs/katello-server-ca.crt"
  SSLVerifyClient         optional
  SSLCACertificateFile    "/etc/pki/katello/certs/katello-default-ca.crt"

  ## Container Gateway
  <Location "/pulpcore_registry/v2/">
    Require expr %{SSL_CLIENT_S_DN_CN} == "centos7-64.example.com"
    RequestHeader set SSL_CLIENT_S_DN "admin"
  </Location>
  ProxyPass /v1/ https://centos7-64.example.com:9090/container_gateway/v1/
  ProxyPassReverse /v1/ https://centos7-64.example.com:9090/container_gateway/v1/
  ProxyPass /v2/ https://centos7-64.example.com:9090/container_gateway/v2/
  ProxyPassReverse /v2/ https://centos7-64.example.com:9090/container_gateway/v2/
</VirtualHost>
```

My main concern is with two entries for `<Location "/pulpcore_registry/v2/">`. I don't know Apache well enough to say whether this works well.

manifests/init.pp
```diff
@@ -254,6 +254,10 @@
...
+    class { 'foreman_proxy::plugin::container_gateway':
+      pulp_endpoint => "https://${servername}",
+    }
```
ekohl, line 260: This is included unconditionally. Is it intended that it's also included when it's shared with a Foreman install?

ianballou: I put this here because I thought the container gateway should always be paired with `pulpcore::plugin::container` on a proxy.  I was assuming if Pulpcore is installed then Katello is in the picture somewhere too.  Is that not the case?

ianballou:
> My main concern is with two entries for <Location "/pulpcore_registry/v2/">. I don't know Apache well enough to say whether this works well.

@ekohl my comment above was in regards to that: 

> I suppose it would be better for all the directives to be under the same <Location "/pulpcore_registry/v2/">, but I believe the two Locations with the same path will just be merged (http://httpd.apache.org/docs/current/sections.html#merging). I tried reversing the order and it doesn't seem to matter either.

I'm guessing there's some way to overwrite/override the previous config, but I wonder if the logic would be messier.  I also like the idea of keeping the container gateway configuration separate just from an organizational standpoint.  It seems like Apache should have no issue with it and order doesn't matter here (from my testing).

manifests/init.pp
ekohl: Naming here is confusing, but we have 2 scenarios:
* Deployed standalone, typically a mirror. In downstream terms this would be a capsule
* Deployed together with Foreman

I think you handle the standalone case, but I'm asking about the latter case. From conversations I seem to recall that the container gateway plugin is only relevant in case there is no Katello serving the API.

You probably also create a nice loop where the container gateway will end up asking itself since `/v2` is proxied to it.

ianballou: Ah right, that slipped my mind. Right now I only want the container gateway to exist in mirrors. Katello handles the registry if the smart proxy exists on the same machine.

Is there a best parameter to check for a mirror? I believe in the pulp plugin "mirror" will be set to true. Perhaps there's a more general parameter?

ekohl: There is `$shared_with_foreman_vhost`.

manifests/init.pp
```diff
@@ -254,6 +254,12 @@
 
   if $enable_docker {
     include pulpcore::plugin::container
+    include foreman_proxy_content::container
+    if !$shared_with_foreman_vhost {
```
ekohl, line 258: This also works:
```suggestion
    unless $shared_with_foreman_vhost {
```

However, i do think it should also make `include foreman_proxy_content::container` dependent on it.

ianballou: I've updated and re-tested.

ekohl (approved):
I think this is correct now. Would you mind squashing the commits before merging though?

ianballou:
@ekohl I've squashed but you'll have to merge for me.

PR #323, opened 2021-02-05
Configure qpid dispatch router on EL8

ehelms:


jturel (approved):
This looks pivotal in terms of katello-agent working on EL8 and I support the change.

ehelms:
Forgot to fix the indent, got it now

jturel:
Are there any tests that need to be updated that might not be making certain assertions based on an el7 check?

ehelms:
I made one small update, but I think things are good as is for this change. I would like to add some additional testing in a follow up because I would like to pursue having dispatch router off by default and users turning on katello-agent support.

jturel:
LGTM

spec/classes/foreman_proxy_content_spec.rb
```diff
@@ -12,12 +12,6 @@
       end
 
       context 'with pulpcore' do
```
ekohl, line 14: This now becomes the same as `without parameters` so we can merge those two groups. https://github.com/theforeman/puppet-foreman_proxy_content/pull/325

PR #327, opened 2021-02-16
Refs #31878: Use client certificate auth to talk from qpid-dispatch t…

ehelms:
…o qpid

### DIFF
--- a/manifests/dispatch_router.pp
+++ b/manifests/dispatch_router.pp
@@ -34,21 +34,15 @@ class foreman_proxy_content::dispatch_router (
 
   contain qpid::router
 
-  # SSL Certificate Configuration
-  class { 'certs::qpid_router':
-    require => Class['qpid::router::install'],
-  }
-  ~> qpid::router::ssl_profile { 'client':
-    ca   => $certs::ca_cert,
-    cert => $certs::qpid_router::client_cert,
-    key  => $certs::qpid_router::client_key,
-  }
-  ~> qpid::router::ssl_profile { 'server':
+  include certs::qpid_router::server
+
+  qpid::router::ssl_profile { 'server':
     ca        => $certs::ca_cert,
-    cert      => $certs::qpid_router::server_cert,
-    key       => $certs::qpid_router::server_key,
+    cert      => $certs::qpid_router::server::cert,
+    key       => $certs::qpid_router::server::key,
     ciphers   => $ssl_ciphers,
     protocols => $ssl_protocols,
+    subscribe => Class['certs::qpid_router::server'],
   }
 
   # Listen for katello-agent clients
--- a/manifests/dispatch_router/connector.pp
+++ b/manifests/dispatch_router/connector.pp
@@ -10,6 +10,15 @@ class foreman_proxy_content::dispatch_router::connector (
   Stdlib::Port $port = 5646,
 ) {
   include foreman_proxy_content::dispatch_router
+  include certs::qpid_router::client
+
+  qpid::router::ssl_profile { 'client':
+    ca        => $certs::ca_cert,
+    cert      => $certs::qpid_router::client::cert,
+    key       => $certs::qpid_router::client::key,
+    notify    => Service['qdrouterd'],
+    subscribe => Class['certs::qpid_router::server'],
+  }
 
   qpid::router::connector { 'hub':
     host         => $host,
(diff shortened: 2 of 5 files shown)


ehelms:
Needed in conjunction with https://github.com/theforeman/puppet-katello/pull/384 -- serves to simplify the configuration and setup more and add an additional layer of security by turning auth on for qpidd.

manifests/dispatch_router/hub.pp
```diff
@@ -11,26 +11,23 @@
...
   Stdlib::Port $broker_port = 5671,
-  String $sasl_mech = 'PLAIN',
-  String $sasl_username = 'katello_agent',
-  String $sasl_password = extlib::cache_data('foreman_cache_data', 'qpid_router_sasl_password', extlib::random_password(16)),
 ) {
   include foreman_proxy_content::dispatch_router
 
+  class { 'certs::qpid_router_client':
+    hostname => 'qpid_router_katello_agent',
+  }
```
ehelms, line 24: Moving the new `qpid_router_client` class declaration inside the hub and connector allow us to configure the generated client certificates differently. For a hub, we want to generate a set of client certificates with a special CN set to `qpid_router_katello_agent` that matches the ACLs set in puppet-katello (https://github.com/theforeman/puppet-katello/pull/384/files#diff-b39fa0a3b1f98c974d4d7d21d4fb0e2d1292dd442ae605d6063c96b9cd9ed932R1). This allows us to be more specific and identify the difference between when Katello and qpid-dispatch connect to qpid. For example:

```
  connection                   cproc      cpid   mech       auth                                                           connected  idle  msgIn  msgOut
  =========================================================================================================================================================
  qpid.[::1]:5671-[::1]:35360  -                 EXTERNAL   qpid_router_katello_agent@QPID                                 6s         4s       0      0
  qpid.[::1]:5671-[::1]:35364  -                 EXTERNAL   pipe-katello-server-nightly-centos7.wareagle.example.com@QPID  2s         0s       0      0
  qpid.[::1]:5671-[::1]:35366  qpid-stat  26105  ANONYMOUS  anonymous@QPID                                                 0s         0s       1      0
```

If both used a client certificate with the hostname, we would not be able to easily tell them apart nor set the ACLs properly. 

On a connector style deployment, such as a proxy with content, we want to deploy the client certificates with the hostname (https://github.com/theforeman/puppet-foreman_proxy_content/pull/327/files#diff-a447b8870e696e6842730621455173b0e03529b5d108cc645e379a45daf2fc30R13) because in that case you have a qpid-dispatch connecting to another qpid-dispatch.

ehelms:
Undertaking this exercise, made me think when we tackle making katello-agent optional and disabled by default we should move the qpid handling out of puppet-katello and into this module to better centralize the connected pieces.

jturel (approved):
ACK - tested working in conjunction w/ my Katello PR

manifests/dispatch_router.pp
```diff
@@ -34,19 +34,12 @@
...
-  class { 'certs::qpid_router':
-    require => Class['qpid::router::install'],
-  }
-  ~> qpid::router::ssl_profile { 'client':
-    ca   => $certs::ca_cert,
-    cert => $certs::qpid_router::client_cert,
-    key  => $certs::qpid_router::client_key,
-  }
-  ~> qpid::router::ssl_profile { 'server':
+  include certs::qpid_router
```
ekohl, line 37: Previously there was a notify that if a certificate changed, qpid would be restarted. Not 100% sure it worked, but after this change it certainly is no longer there. While I prefer the `include` over `class { '...': }`, I'd ask you to add `subscribe => Class['certs::qpid_router']` to the `ssl_profile`. That way you also guarantee that the certs that the profile uses are present prior to starting qpid router.

manifests/dispatch_router/connector.pp
```diff
@@ -10,6 +10,13 @@
   Stdlib::Port $port = 5646,
 ) {
   include foreman_proxy_content::dispatch_router
+  include certs::qpid_router_client
```
ekohl, line 13: Same as above here.

ekohl (changes requested):
This will require a new version of the certs module. We should have a matching update in metadata.json so we know at a module level they are compatible.

jlsherrill (approved):

PR #328, opened 2021-02-18
Fixes #31868: Switch to /pulp/content/ by default for clients

ehelms:
With Pulp 3 present, clients should start talking to /pulp/content/
to eventually be able to drop the old Pulp 2 URLs in our reverse
proxy configuration.

### DIFF
--- a/spec/acceptance/bootstrap_rpm_spec.rb
+++ b/spec/acceptance/bootstrap_rpm_spec.rb
@@ -66,7 +66,7 @@ describe 'bootstrap_rpm', :order => :defined do
       its(:content) { should match /full_refresh_on_yum = 1/ }
       its(:content) { should match /package_profile_on_trans = 1/ }
       its(:content) { should match /hostname = #{host_inventory['fqdn']}/ }
-      its(:content) { should match /baseurl = https:\/\/#{host_inventory['fqdn']}\/pulp\/repos/ }
+      its(:content) { should match %r{baseurl = https://#{host_inventory['fqdn']}/pulp/content/} }
       its(:content) { should match /port = 443/ }
     end
   end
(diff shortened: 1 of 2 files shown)


spec/acceptance/bootstrap_rpm_spec.rb
```diff
@@ -66,7 +66,7 @@
       its(:content) { should match /full_refresh_on_yum = 1/ }
       its(:content) { should match /package_profile_on_trans = 1/ }
       its(:content) { should match /hostname = #{host_inventory['fqdn']}/ }
-      its(:content) { should match /baseurl = https:\/\/#{host_inventory['fqdn']}\/pulp\/repos/ }
+      its(:content) { should match /baseurl = https:\/\/#{host_inventory['fqdn']}\/pulp\/content/ }
```
ekohl, line 69: I always like to use `%r{}` when the regex contains slashes.
```suggestion
      its(:content) { should match %r{baseurl = https://#{host_inventory['fqdn']}/pulp/content/} }
```

(same below)

ekohl (commented):
Out of interest (and I haven't looked into this), is this URL provided in the ENC for a host? In other words, in the long term can registration or some API client actually discover this location?

ehelms:
> Out of interest (and I haven't looked into this), is this URL provided in the ENC for a host? In other words, in the long term can registration or some API client actually discover this location?

This configuration is going to be handled by global registration in the long run.

ekohl:
> This configuration is going to be handled by global registration in the long run.

That's sort of what I'm getting at. How much of the (Foreman internal) infra is already in place?

ehelms:
> That's sort of what I'm getting at. How much of the (Foreman internal) infra is already in place?

None yet, it is on the Roadmap. I moved this out of draft after chatting with @jlsherrill -- seems like we could make this simple change for 4.0 and start the process of getting clients using it to drop the old paths sooner than if we waited for 4.1.

ekohl (approved):

PR #329, opened 2021-02-19
Refs #32037: Add flag to enable katello_agent infrastructure and disable it by default

ehelms:


ehelms:
See https://github.com/theforeman/puppet-katello/pull/387 for more details

manifests/init.pp
```diff
@@ -146,7 +147,7 @@
 
   include foreman_proxy_content::pub_dir
 
-  if $qpid_router {
+  if $enable_katello_agent {
```
ekohl, line 150: I'm wondering if we should implement an `else` for this where we ensure it's stopped.

ehelms: Will we need to add some additional parameters to puppet-qpid to be able to ensure things are stopped? Given https://github.com/theforeman/puppet-qpid/blob/master/manifests/router/service.pp#L6

ehelms:
Relies on https://github.com/theforeman/puppet-qpid/pull/148

manifests/init.pp
```diff
@@ -146,18 +147,20 @@
...
+    agent_port     => $qpid_router_agent_port,
+    ssl_ciphers    => $qpid_router_ssl_ciphers,
+    ssl_protocols  => $qpid_router_ssl_protocols,
+    logging_level  => $qpid_router_logging_level,
+    logging        => $qpid_router_logging,
+    logging_path   => $qpid_router_logging_path,
+    service_ensure => $enable_katello_agent,
+    service_enable => $enable_katello_agent,
+  }
+  contain foreman_proxy_content::dispatch_router
```
ekohl, line 161: The only concern I have here is that this now always installs the router packages, even if the service is then disabled. Previously it didn't even install the packages.

wbclark:
This needs a puppet-qpid release ( https://github.com/theforeman/puppet-qpid/pull/151 ) to properly test it

Once that happens it also needs the dependency bumped to get the version which added that feature

ekohl:
> This needs a puppet-qpid release ( [theforeman/puppet-qpid#151](https://github.com/theforeman/puppet-qpid/pull/151) ) to properly test it

In the past I've changed `.fixtures.yml` to point to my branch just so CI would run. Obviously you need to update it prior to merging with a dependency on actual change but it's a good trick to remember.

manifests/init.pp
wbclark: Do we actually gain anything by having

```
class { 'foreman_proxy_content::dispatch_router':
  ...
}
contain foreman_proxy_content::dispatch_router
```

Declared outside of the

```
if $enable_katello_agent {
  ...
}
```

block?

It's not clear to me why you are moving it here.

manifests/dispatch_router.pp
```diff
@@ -30,9 +34,13 @@
...
   Enum['file', 'syslog'] $logging = 'syslog',
   Stdlib::Absolutepath $logging_path = '/var/log/qdrouterd',
+
+  Boolean $service_ensure = false,
 ) {
 
-  contain qpid::router
+  class { 'qpid::router':
+    service_ensure => $service_ensure,
+  }
```
ekohl, line 43: Please keep `contain`
```suggestion
  class { 'qpid::router':
    service_ensure => $service_ensure,
  }
  contain qpid::router
```

This ensures correct chaining in the dependency graph. It ensures that all resources from `qpid::router` are contained within this class. It's roughly short form for:
```puppet
include qpid::router
Class['qpid::router'] -> Class['foreman_proxy_content::dispatch_router']
```

Not exactly (because what I just wrote is `require qpid::router`), but I'm not sure how else to explain it.

manifests/init.pp
```diff
@@ -155,18 +156,19 @@
...
-      agent_addr    => $qpid_router_agent_addr,
-      agent_port    => $qpid_router_agent_port,
-      ssl_ciphers   => $qpid_router_ssl_ciphers,
-      ssl_protocols => $qpid_router_ssl_protocols,
-      logging_level => $qpid_router_logging_level,
-      logging       => $qpid_router_logging,
-      logging_path  => $qpid_router_logging_path,
-    }
-    contain foreman_proxy_content::dispatch_router
+  class { 'foreman_proxy_content::dispatch_router':
```
ekohl, line 159: My concern hasn't been addressed: previously it was possible to not even install qpid router at all. Now it's always installed and configured, just disabled.

Perhaps we need to implement this:
```suggestion
  class { 'foreman_proxy_content::dispatch_router':
    ensure => bool2str($enable_katello_agent, 'present', 'absent'),
```

manifests/init.pp
ehelms: It exists outside of the `enable_katello_agent` block because we don't know the previous state, and therefore we want to ensure the service isn't running if it previously was running. If we were to always put it within the `enable_katello_agent` block you could enable it and then disable it while leaving the services running.

The question I posed in a different forum, which @ekohl asks here (https://github.com/theforeman/puppet-foreman_proxy_content/pull/329#discussion_r605534618) is whether it is sufficient to just disable the services when disabling katello-agent support (or when freshly setting things up to have no presence of) or if we need to propagate this effect to all entities: services, packages, config files. The same question applies here: https://github.com/theforeman/puppet-katello/pull/387/files#diff-4d995f8e71429e47718f73b6c9d8cad415c63a67017803143f5fa1a929d0f730R19

manifests/init.pp
ehelms: Take a look at https://github.com/theforeman/puppet-qpid/pull/154

wbclark (commented):
The failure on `bundle exec rake validate` is happening because of a typo'd filename for the dispatch router acceptance test

I confirmed that was resolved with `git mv spec/acceptance/dispatch_router_spec.{pp,rb}`

ehelms:
> The failure on `bundle exec rake validate` is happening because of a typo'd filename for the dispatch router acceptance test
> 
> I confirmed that was resolved with `git mv spec/acceptance/dispatch_router_spec.{pp,rb}`

Thanks. Things were not failing locally as it didnt seem to care about the extension and I could not see the forest for the trees.

manifests/dispatch_router.pp
```diff
@@ -30,8 +34,13 @@
...
+
+  Enum['present', 'absent'] $ensure = 'present',
 ) {
 
+  class { 'qpid::router':
+    ensure => $ensure,
+  }
   contain qpid::router
 
   include certs::qpid_router::server
```
ekohl, line 46: Perhaps this should be inside `if $ensure == 'present' {}`. However, that also implies the `ssl_profile` should be hidden. That also makes me wonder if all router statements (listener and log) should be hidden.

manifests/init.pp
```diff
@@ -22,8 +24,6 @@
 #
 # $reverse_proxy_port::                        Reverse proxy listening port
 #
-# $qpid_router::                               Configure qpid dispatch router
```
ekohl, line 25: Note to self: this may need an installer migration.

metadata.json
```diff
@@ -26,7 +26,7 @@
     },
     {
       "name": "katello/qpid",
-      "version_requirement": ">= 4.5.0 < 8.0.0"
+      "version_requirement": ">= 7.1.0 < 8.0.0"
```
ekohl, line 29: Same note here as Katello: 7.1.0 doesn't contain the required patches.

ekohl (commented):
Both this and https://github.com/theforeman/puppet-katello/pull/387 use `Fixes #32037`. To me it's a bit confusing to have 2 PRs both close a Redmine issue. Since we'll likely need an installer migration to really finish it, should all of them be `Refs` and only the migration would use `Fixes`?

ehelms:
> Both this and [theforeman/puppet-katello#387](https://github.com/theforeman/puppet-katello/pull/387) use `Fixes #32037`. To me it's a bit confusing to have 2 PRs both close a Redmine issue. Since we'll likely need an installer migration to really finish it, should all of them be `Refs` and only the migration would use `Fixes`?

Will do.

manifests/init.pp
```diff
@@ -155,18 +156,19 @@
...
-    contain foreman_proxy_content::dispatch_router
+  class { 'foreman_proxy_content::dispatch_router':
+    agent_addr    => $qpid_router_agent_addr,
+    agent_port    => $qpid_router_agent_port,
+    ssl_ciphers   => $qpid_router_ssl_ciphers,
+    ssl_protocols => $qpid_router_ssl_protocols,
+    logging_level => $qpid_router_logging_level,
+    logging       => $qpid_router_logging,
+    logging_path  => $qpid_router_logging_path,
+    ensure        => bool2str($enable_katello_agent, 'present', 'absent'),
```
ekohl, line 167: Odd: I thought there was a linting rule that preferred `ensure` to be the first line if given.

manifests/dispatch_router.pp
```diff
@@ -30,43 +34,50 @@
   String $logging_level = 'info+',
   Enum['file', 'syslog'] $logging = 'syslog',
   Stdlib::Absolutepath $logging_path = '/var/log/qdrouterd',
+
+  Enum['present', 'absent'] $ensure = 'present',
```
ekohl, line 38: It is common to have `$ensure` as the first

ekohl (approved):

manifests/init.pp
ekohl: Done in https://github.com/theforeman/foreman-installer/pull/675

ekohl (approved):

PR #330, opened 2021-02-23
Update Github Actions workflow

ekohl:
This applies the lessons learned while applying this to Voxpupuli and The Foreman orgs.


ekohl:
Unit tests are broken since pulpcore is broken with the git version of camptocamp/systemd.

PR #331, opened 2021-02-26
Add pulpcore_django_secret_key parameter

laugmanuel:
Currently, the pulpcore parameter `django_secret_key` is generated/cached using `extlib` inside `puppet-pulpcore`. As the parameter is not exposed here, this breaks when using multiple puppetservers.

With this PR a user can explicitly specify that parameter. If the parameter is not set, it will still use the implementation of `puppet-pulpcore` (to not break existing installs).

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -113,7 +113,7 @@ class foreman_proxy_content (
   Stdlib::Absolutepath $pulpcore_postgresql_ssl_key = '/etc/pki/katello/private/pulpcore-database.key',
   Stdlib::Absolutepath $pulpcore_postgresql_ssl_root_ca = '/etc/pki/tls/certs/ca-bundle.crt',
   Integer[0] $pulpcore_worker_count = $foreman_proxy_content::params::pulpcore_worker_count,
-  Optional[String] $pulpcore_django_secret_key = undef,
+  Optional[String[50]] $pulpcore_django_secret_key = undef,
 ) inherits foreman_proxy_content::params {
   include certs
   include foreman_proxy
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -111,6 +111,7 @@
   Stdlib::Absolutepath $pulpcore_postgresql_ssl_key = '/etc/pki/katello/private/pulpcore-database.key',
   Stdlib::Absolutepath $pulpcore_postgresql_ssl_root_ca = '/etc/pki/tls/certs/ca-bundle.crt',
   Integer[0] $pulpcore_worker_count = $foreman_proxy_content::params::pulpcore_worker_count,
+  Optional[String] $pulpcore_django_secret_key = $foreman_proxy_content::params::pulpcore_django_secret_key,
```
ekohl, line 114: We've started to move static parameters to `init.pp` where possible
```suggestion
  Optional[String] $pulpcore_django_secret_key = undef,
```

ekohl (changes requested):
Could you also add a parameter description for this?

laugmanuel:
> Could you also add a parameter description for this?

Seems like I always forget the same thing 🤦🏻 I've added a description for the parameter.

manifests/init.pp
```diff
@@ -111,6 +113,7 @@
   Stdlib::Absolutepath $pulpcore_postgresql_ssl_key = '/etc/pki/katello/private/pulpcore-database.key',
   Stdlib::Absolutepath $pulpcore_postgresql_ssl_root_ca = '/etc/pki/tls/certs/ca-bundle.crt',
   Integer[0] $pulpcore_worker_count = $foreman_proxy_content::params::pulpcore_worker_count,
+  Optional[String] $pulpcore_django_secret_key = undef,
```
ekohl, line 116: It is recommended to use at least 50 characters. Should we enforce that?
```suggestion
  Optional[String[50]] $pulpcore_django_secret_key = undef,
```

ekohl (commented):
Overall :+1:

ekohl (approved):

PR #333, opened 2021-03-08
Support Puppet 7

ekohl:


wbclark:
Re-running tests as I expect the failures we saw would be resolved by https://github.com/theforeman/puppet-foreman_proxy_content/commit/069f9a59270657d50da32c214200cf831953cecc

wbclark:
Unfortunately the issue is in puppet-redis fixture:

```
Evaluation Error: Error while evaluating a Function Call, Could not find class ::systemd::systemctl::daemon_reload ... (file: /home/runner/work/puppet-foreman_proxy_content/puppet-foreman_proxy_content/spec/fixtures/modules/redis/manifests/ulimit.pp, line: 54, column: 5)
```

PR #335, opened 2021-03-21
Fixes #32149 - Expose pulpcore-{content,api} gunicorn worker timeouts

wbclark:


manifests/init.pp
```diff
@@ -4,74 +4,78 @@
...
-#                                       Enabling more than 8 workers, even with additional CPU cores available, likely results in performance
-#                                       degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should be done
-#                                       incrementally with benchmarking at each step to determine an optimal value for your deployment.
+# $pulpcore_worker_count::                     Number of pulpcore workers. Defaults to 8 or the number of CPU cores, whichever is smaller.
+#                                              Enabling more than 8 workers, even with additional CPU cores available, likely results in performance
+#                                              degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should be done
+#                                              incrementally with benchmarking at each step to determine an optimal value for your deployment.
+#
+# $pulpcore_content_gunicorn_worker_timeout::  Gunicorn worker timeout in seconds for the pulpcore-content.service
```
ehelms, line 76: In the pulpcore module we use `service` instead of `gunicorn`, I think I like the use of `service` better and its provides consistency with the base module.

.fixtures.yml
```diff
@@ -26,7 +26,9 @@ fixtures:
     selinux_core:
       repo: "https://github.com/puppetlabs/puppetlabs-selinux_core"
       puppet_version: ">= 6.0.0"
-    systemd:       "https://github.com/camptocamp/puppet-systemd.git"
+    systemd:
+      repo: 'https://github.com/camptocamp/puppet-systemd'
+      ref: '2.12.0'
```
ehelms, line 31: It'd be great if you could open a PR with just this change in it, I am needing locally for a few PRs myself.

wbclark: Done at https://github.com/theforeman/puppet-foreman_proxy_content/pull/336

ehelms: Thanks!

ehelms:
Needs a rebase with the systemd fix now in

ekohl (changes requested):
Can you update `metadata.json` to require at least 3.2.0 since that introduces the parameters?

wbclark:
Thanks, I've pushed an update which addressed all feedback

ehelms (approved):

PR #336, opened 2021-03-22
Pin camptocamp/systemd to 2.12.0 in testing

wbclark:

PR #337, opened 2021-03-22
Fixes #32161: Set disablereuse=on and retry=0 on reverse proxy

ehelms:
This change aims to avoid timeouts and delays when client connection
is interrupted reverse proxying back to the Foreman server.
The reverse proxy is meant for communication back to primary Foreman
server where there are not multiple "workers" to receive a request.
This should reduce the number of clients getting 502 proxy errors
and having to be retried manually thus increasing the reliability.


ehelms:
Currently includes https://github.com/theforeman/puppet-foreman_proxy_content/pull/336 to get tests to pass. I will remove that change and rebase it once the other PR is in.

ekohl (approved):

PR #339, opened 2021-03-31
Refs #32239 - Drop pulp 2 parameters from foreman_proxy::plugin::pulp

ehelms:
Requires https://github.com/theforeman/puppet-foreman_proxy/pull/638 be merged first

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -277,8 +277,6 @@ class foreman_proxy_content (
   include pulpcore::plugin::certguard # Required to be present by Katello when syncing a content proxy
 
   class { 'foreman_proxy::plugin::pulp':
-    enabled              => false,
-    pulpnode_enabled     => false,
     pulpcore_enabled     => true,
     pulpcore_mirror      => $pulpcore_mirror,
     pulpcore_api_url     => "https://${servername}",
--- a/metadata.json
+++ b/metadata.json
@@ -22,7 +22,7 @@
     },
     {
       "name": "theforeman/foreman_proxy",
-      "version_requirement": ">= 17.1.0 < 18.0.0"
+      "version_requirement": ">= 18.0.0 < 19.0.0"
     },
     {
       "name": "katello/qpid",


wbclark (commented):
Tests are green but that is a little misleading, as until https://github.com/theforeman/puppet-foreman_proxy/pull/638 gets merged, dropping these params here means we are getting the defaults defined in `foreman_proxy::plugin::pulp::params` which means actually the tests ran with `enabled => true`.

The fact that that didn't break any tests seems to be a reflection of the fact that pulp2 tests were already dropped from from `foreman_proxy_content_spec.rb`. Maybe it would have been better if we had instead had a test that pulp2 should really be disabled by default, and then we'd get the red tests we should expect to see here by dropping the parameters and falling back to the defaults defined in the other module.

Anyway, adding such a test now wouldn't really make sense, so all of this is to say: let's make sure to re-run tests here once https://github.com/theforeman/puppet-foreman_proxy/pull/638 is actually merged :)

ekohl (commented):
I wonder if this should depend on the version of theforeman/foreman_proxy that includes https://github.com/theforeman/puppet-foreman_proxy/commit/2a500f4dae38a97b0ffa65de729a13dbfa9ddd18. We need to major bump that module anyway and that means updating metadata.json anyway.

ehelms:
> I wonder if this should depend on the version of theforeman/foreman_proxy that includes [theforeman/puppet-foreman_proxy@2a500f4](https://github.com/theforeman/puppet-foreman_proxy/commit/2a500f4dae38a97b0ffa65de729a13dbfa9ddd18). We need to major bump that module anyway and that means updating metadata.json anyway.

In theory yes, due to the behavior difference @wbclark pointed out. Due to parameters, pipelines will fail as these parameters being removed here are not present. If w can do that major bump now, then I'd be up for including it here. If we need to wait on the major version bump, then we should merge this as is to prevent pipeline and testing breakages.

wbclark (commented):
:+1:

> I wonder if this should depend on the version of theforeman/foreman_proxy that includes [theforeman/puppet-foreman_proxy@2a500f4](https://github.com/theforeman/puppet-foreman_proxy/commit/2a500f4dae38a97b0ffa65de729a13dbfa9ddd18). We need to major bump that module anyway and that means updating metadata.json anyway.

In some technical sense it doesn't actually depend on that commit because we will never pass those params and `puppet::foreman_proxy` either wouldn't accept them anyway (as intended), or for older versions would fall back to defaults. But the intention here is clearly that we should not install pulp2, and I think that's a strong argument for bumping the `version_requirement`.

I can start things off on that major version bump for `puppet-foreman_proxy`

ehelms:
> I can start things off on that major version bump for `puppet-foreman_proxy`

I jumped on this -- https://github.com/theforeman/puppet-foreman_proxy/pull/659
I realized I could just do it just as fast as typing :)

ekohl (approved):

PR #342, opened 2021-04-15
Fixes #32338 - expose pulpcore allowed_content_checksums

jlsherrill:

PR #343, opened 2021-04-20
Fixes #32160 - enable ansible content

jlsherrill:


ekohl:
We'll need to update metadata.json as well to depend on a version that includes the new code. That probably requires https://github.com/theforeman/puppet-pulpcore/pull/184 to be merged and a version bump, at least in git. After that I also think the tests should pass.

jlsherrill:
yes, i will likely bump the requires in https://github.com/theforeman/puppet-foreman_proxy_content/pull/342  first once the new version is released and then re-test this one

manifests/init.pp
```diff
@@ -12,6 +12,8 @@
 #
 # $enable_deb::                                Enable debian content plugin
 #
+# $enable_ansible::                            Enable ansible content plugin
```
ehelms, line 15: Is the discrepancy in naming intentional? In the Katello application itself, the setting is `enable_ansible_collection` (https://github.com/theforeman/puppet-katello/pull/401/files#diff-510884a7414666a0d988133c5f869510210d883aa1580d1248e74f1d16ca7d91R13)

jlsherrill: yes, this is intentional.  the pulp plugin is called 'ansible', but provides two different types of 'repositories' (really remotes/distributors), ansible_collections and ansible_roles.  Katello only supports ansible_collections at this time, and would likely implement roles as a different repo type (if we ever implement support)

ehelms:
You can now update https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/metadata.json#L37 to a minimum of `3.3.0`

jlsherrill:
will rebase once https://github.com/theforeman/puppet-foreman_proxy_content/pull/342 goes in

jlsherrill:
went ahead and rebased this and updated metadata.json, so which ever goes in first is fine

manifests/init.pp
wbclark, line 15:
```suggestion
# $enable_ansible::                            Enable ansible content plugin to sync and serve ansible content
```

I had this feedback originally on the puppet-katello PR but it makes more sense here. The reason is that this is what users will see when looking at the output of `foreman-installer --help`

The implementation you have is consistent with how the other content plugin params are already documented. I think there is a potential clarity issue with ansible content specifically because we support many other features with ansible (REX, adding roles to hosts/hostgroups, we provide ansible collections to automate operations and configuration of Foreman/Katello).

If we think there is value in it (keeping the doc strings consistent for all of the content plugin params), I can tackle this in a separate PR which updates the other content plugin param docs as well. I don't care how it gets done, I just want to make sure it's crystal clear for users what this does.

wbclark: In fact I will go ahead and do that, so don't worry about it here :)

ehelms (approved):

PR #344, opened 2021-04-22
Improve clarity of content type feature param documentation strings

wbclark:
This is based on #343 

I can rebase once that is merged if it is needed

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -4,15 +4,15 @@
 #
 # === Parameters:
 #
-# $enable_yum::                                Enable rpm content plugin, including syncing of yum content
+# $enable_yum::                                Enable the RPM content feature. This allows syncing, managing and serving RPM content to be consumed by package managers like yum and dnf.
 #
-# $enable_file::                               Enable file content plugin
+# $enable_file::                               Enable the file content feature. This allows syncing, managing, and serving file content.
 #
-# $enable_docker::                             Enable docker content plugin
+# $enable_docker::                             Enable the container content feature. This allows syncing, managing, and serving container content.
 #
-# $enable_deb::                                Enable debian content plugin
+# $enable_deb::                                Enable the Debian content feature. This allows syncing, managing, and serving Debian content.
 #
-# $enable_ansible::                            Enable ansible content plugin
+# $enable_ansible::                            Enable the Ansible content feature. This allows syncing, managing, and serving Ansible content.
 #
 # $enable_katello_agent::                      Enable katello-agent for remote yum actions
 #


wbclark:
Rebased

manifests/init.pp
```diff
@@ -4,15 +4,15 @@
 #
 # === Parameters:
 #
-# $enable_yum::                                Enable rpm content plugin, including syncing of yum content
+# $enable_yum::                                Enable rpm content plugin to sync and serve rpm content
```
ekohl, line 7:
```suggestion
# $enable_yum::                                Enable rpm content plugin to sync and serve rpm content to be consumed by package managers like yum and dnf
```

manifests/init.pp
```diff
@@ -4,15 +4,15 @@
...
 #
-# $enable_file::                               Enable file content plugin
+# $enable_file::                               Enable file content plugin to sync and serve file content
 #
-# $enable_docker::                             Enable docker content plugin
+# $enable_docker::                             Enable docker content plugin to sync and serve docker content
```
ekohl, line 11: Technically this is the container plugin, we just haven't renamed it yet.
```suggestion
# $enable_docker::                             Enable container content plugin to sync and serve container content
```

manifests/init.pp
```diff
@@ -4,15 +4,15 @@
...
 #
-# $enable_deb::                                Enable debian content plugin
+# $enable_deb::                                Enable debian content plugin to sync and serve debian content
```
ekohl, line 13: We should capitalize
```suggestion
# $enable_deb::                                Enable Debian content plugin to sync and serve Debian content
```

manifests/init.pp
```diff
@@ -4,15 +4,15 @@
...
 #
-# $enable_ansible::                            Enable ansible content plugin
+# $enable_ansible::                            Enable ansible content plugin to sync and serve ansible content
```
ekohl, line 15:
```suggestion
# $enable_ansible::                            Enable Ansible content plugin to sync and serve Ansible content
```

ekohl (changes requested):

ehelms:
Instead of `sync and serve`, consider `manage` as you can are enabling the content to do a lot more than just sync and serve it.

wbclark:
> Instead of `sync and serve`, consider `manage` as you can are enabling the content to do a lot more than just sync and serve it.

I am torn because `manage` might be more technically correct, but I think the specificity of `sync and serve` is probably clearer and more immediately intuitive to a larger set of users.

I think I prefer either `sync and serve` or maybe `sync, manage, and serve` for that reason. What do you think @ehelms ?

wbclark:
Thanks for the suggestions @ekohl . I've made the update based on your recommendations.

I kept the `sync and serve` language for now. Definitely open to further discussion on that.

manifests/init.pp
```diff
@@ -4,15 +4,15 @@
 #
 # === Parameters:
 #
-# $enable_yum::                                Enable rpm content plugin, including syncing of yum content
+# $enable_yum::                                Enable rpm content plugin to sync and serve rpm content to be consumed by package manager like yum and dnf
```
ehelms, line 7:
```suggestion
# $enable_yum::                                Enable the RPM content plugin to sync and serve rpm content to be consumed by package manager like yum and dnf
```

If we are going to refer to `content plugin` I think we should then also include that it is `Pulp RPM content plugin` to be specific here. Otherwise drop the `content plugin` and just refer to it as `Enable RPM content`

ekohl: I would also think about `Enable support for RPM content`

ehelms: > I would also think about `Enable support for RPM content`

I like this wording

wbclark: How about:

`Enable feature to sync, manage, and serve RPM content to be consumed by package managers like yum and dnf.`

`Enable feature to sync, manage, and serve file content.`

et cetera

IMO it's the best balance yet of being truly accurate (we're enabling the entire feature, not only the one piece of that feature which is the pulp plugin) and also very easily understood (you can sync, manage, and serve X content type)

wbclark: BTW I used `enable feature` instead of `enable support` because otherwise I promise there is SOMEBODY out there who is going to think it has to do with SLAs :)

manifests/init.pp
```diff
@@ -4,15 +4,15 @@
 #
 # === Parameters:
 #
-# $enable_yum::                                Enable rpm content plugin, including syncing of yum content
+# $enable_yum::                                Enable feature to sync, manage, and serve RPM content to be consumed by package managers like yum and dnf.
```
ekohl, line 7: What I liked about `Enable RPM content support` was that the relevant part was near the front. Now as a user you need to "skip" over half a sentence for the relevant bit. If you want to avoid support, you can try this:
```suggestion
# $enable_yum::                                Enable the RPM content feature. This allows syncing, managing and serving RPM content to be consumed by package managers like yum and dnf.
```

wbclark: Great points. This looks good to me and I'll make the change.

manifests/init.pp
```diff
@@ -4,15 +4,15 @@
...
 # === Parameters:
 #
-# $enable_yum::                                Enable rpm content plugin, including syncing of yum content
+# $enable_yum::                                Enable the RPM content feature. This allows syncing, managing and serving RPM content to be consumed by package managers like yum and dnf.
 #
-# $enable_file::                               Enable file content plugin
+# $enable_file::                               Enable the file content feature. This allows syncing, managing, and serving file content.
 #
-# $enable_docker::                             Enable docker content plugin
+# $enable_docker::                             Enable the container contant feature. This allows syncing, managing, and serving container content.
```
ekohl, line 11:
```suggestion
# $enable_docker::                             Enable the container content feature. This allows syncing, managing, and serving container content.
```

ekohl (approved):
Other than the typo, :+1:

PR #347, opened 2021-04-26
Set filedigest to still be compatible with SLES11

sbernhard:
Previously, in the katello-certs-gen-rpm script, the filedigest
algorithm was set like this:

 --define '_source_filedigest_algorithm md5'
 --define '_binary_filedigest_algorithm md5'

### DIFF
--- a/lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
+++ b/lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
@@ -160,6 +160,9 @@ Puppet::Type.type(:bootstrap_rpm).provide(:bootstrap_rpm) do
 
   def spec
     <<~HEREDOC
+      %global _binary_filedigest_algorithm md5
+      %global _source_filedigest_algorithm md5
+
       Name:      %{name}
       Version:   1.0
       Release:   %{release}


ekohl (commented):
Doesn't this break generation of RPMs for FIPS enabled systems?

sbernhard:
Well, I don't know. The RPM dependencies are the same as with foreman 2.1

ehelms:
How is the filedigest set now?

sbernhard:
See: https://access.redhat.com/solutions/131663

Available modes: 
1 = md5
2 = sha1
8 = sha256

Without the PR:

[root@or pub]# rpm -qp --requires katello-ca-consumer-or.master.dev.atix-1.0-1.noarch.rpm
/bin/bash
/bin/sh
/bin/sh
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
subscription-manager
rpmlib(PayloadIsXz) <= 5.2-1

  <rpmTag name="Filedigests">
	<string>c506579077de7b3462fdb4c18a4c41fa33006b47b0462d93a60ff5651a37b168</string>
  </rpmTag>

=> looks like sha256

  <rpmTag name="Filedigestalgo">
	<integer>8</integer>
  </rpmTag>
=> sha256!

With the PR:
[root@or pub]# rpm -qp --requires katello-ca-consumer-or.master.dev.atix-1.0-2.noarch.rpm
/bin/bash
/bin/sh
/bin/sh
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
subscription-manager
rpmlib(PayloadIsXz) <= 5.2-1


  <rpmTag name="Filedigests">
	<string>9bc3b9f3985ccf5ae0c44815de15ed32</string>
  </rpmTag>
=> looks like md5

Filedigestalgo not specified => default!?

ekohl (changes requested):
The problem is that machines in FIPS mode don't accept md5. That means we have to choose between being current and supporting legacy.

I think this can only be accepted if it's hidden behind a parameter that defaults to a current hashing algorithm.

I also missed the info which OS you were on so I checked on CentOS 7:
```console
# grep filedigest /usr/lib/rpm/redhat/macros
%_source_filedigest_algorithm 8
%_binary_filedigest_algorithm 8
```

sbernhard:
Well, in foreman 2.1 the default behavior was md5. See https://github.com/Katello/katello-certs-tools/blob/bf7547467a9f742bbc27ad1670b866f21ff126fd/katello-certs-gen-rpm#L343

So, after the migration to puppet-certs, the default behavior changed from md5 to sha256 with the result that it doesn't work any longer for old legacy systems like sles11 (and maybe rhel5?).

ehelms:
Given SLES11 is in that old legacy bucket, could you opt to use the script directly instead of the RPM? e.g. https://<test>/pub/katello-rhsm-consumer

Otherwise, I lean towards a parameter being added that allows you to pass it along to the provider and include those macros if the parameter is true so that we stick to the current method by default but allows you to override it if you need to generate the consumer RPM with md5.

ekohl:
We're deprecating the RPM and there is now the new registration workflow which doesn't rely on the RPM. I'd recommend that instead.

PR #350, opened 2021-04-28
Fixes #32383: Set Pulp to expect Foreman host as the client authentic…

ehelms:
…ating

This sets Apache to expect the client certificate to contain as the
common name the hostname of Foreman. This corresponds to using the
Foreman client certificates to talk to Pulp's API which is expected
to contain the hostname of Foreman in the certificate.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -226,6 +226,10 @@ class foreman_proxy_content (
     $apache_https_chain = $certs::katello_server_ca_cert
   }
 
+  $api_client_auth_cn_map = Hash($foreman_proxy::trusted_hosts.map |$host| {
+      [$host, 'admin']
+  })
+
   class { 'pulpcore':
     allowed_content_checksums      => $pulpcore_allowed_content_checksums,
     allowed_import_path            => $pulpcore_allowed_import_path,
@@ -254,6 +258,7 @@ class foreman_proxy_content (
     django_secret_key              => $pulpcore_django_secret_key,
     content_service_worker_timeout => $pulpcore_content_service_worker_timeout,
     api_service_worker_timeout     => $pulpcore_api_service_worker_timeout,
+    api_client_auth_cn_map         => $api_client_auth_cn_map,
     before                         => Class['foreman_proxy::plugin::pulp'],
   }
 
--- a/metadata.json
+++ b/metadata.json
@@ -34,7 +34,7 @@
     },
     {
       "name": "theforeman/pulpcore",
-      "version_requirement": ">= 3.3.0 < 4.0.0"
+      "version_requirement": ">= 3.4.0 < 4.0.0"
     }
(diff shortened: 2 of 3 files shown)


ehelms:
Will require https://github.com/theforeman/puppet-pulpcore/pull/186 in a release

ekohl:
I've converted it to a draft to indicate it's blocked on the other PR.

ehelms:
Moved out of draft mode, with the way this functionality works in puppet-pulpcore, the addition of this parameter here is an enhancement and does not break current implementations that us the certificates with CN set to admin. The only thing this needs is a release of or version bump of puppet-pulpcore.

manifests/init.pp
```diff
@@ -255,6 +255,7 @@
     django_secret_key              => $pulpcore_django_secret_key,
     content_service_worker_timeout => $pulpcore_content_service_worker_timeout,
     api_service_worker_timeout     => $pulpcore_api_service_worker_timeout,
+    api_client_auth_cn             => {"${foreman_host}" => 'admin'},
```
ekohl, line 258:
```suggestion
    api_client_auth_cn             => {$foreman_host => 'admin'},
```

ekohl: Reading more, there's no guarantee that the CN on the foreman host is the same as the client cert. I'm trying to decouple that. For example, it may be `foreman.example.com` but the host uses `foreman01.example.com` as the hostname. I'm not entirely sure what the right approach is today, but perhaps the Foreman Proxy's trusted host settings is a better indicator. That typically includes it.

ehelms: The trusted hosts can be a list right? So are suggesting crafting a mapping of each trusted host entry to admin?

ekohl: I'm not entirely sure. That's what I also thought about. I think it makes sense to treat everything that can connect to a Smart Proxy with full permissions also as something that can connect to the Pulp instance with full permissions.

ekohl (approved):

manifests/init.pp
```diff
@@ -226,6 +226,10 @@
     $apache_https_chain = $certs::katello_server_ca_cert
   }
 
+  $api_client_auth_cn_map = Hash($foreman_proxy::trusted_hosts.map |String $host| {
```
ekohl, line 229: I think you're not required to use a data type so this also works.
```suggestion
  $api_client_auth_cn_map = Hash($foreman_proxy::trusted_hosts.map |$host| {
```

However, it's fine to keep it in.

ekohl (approved):

ekohl:
This is safe to merge now, right? All dependencies are in and this should be in before Katello's change is merged. It's also in addition to the current certs so doesn't break. If I'm right about that, please merge when you feel OK about it.

PR #355, opened 2021-05-05
Fixes #32529: Find latest bootstrap RPM by sorting as intenger

ehelms:
The code that finds the maximum (latest) bootstrap RPM was doing
so via a lexical sort which resulted in incorrect ordering for
what should be latest and thus the 9th RPM was always latest. For
example, once the 10th RPM was generated the order went: 1,10,2,3 etc.

This change moves to parsing out the digits (in case there are alphabet
characters present from the dist) in the release, converting to
integers and then sorting.


evgeni (approved):

ekohl (approved):
Should we write an acceptance test for this?

ehelms:
I added an acceptance test that loops and forces re-creation of the RPM repository and then checks that things are updated properly.

ekohl (approved):
Fixing the typo in the commit message on squash.

PR #356, opened 2021-05-06
Fixes #32553 - Remove md5 from allowed content types

jlsherrill:
Previously we had added md5 to maintain md5 across upgrades
such that the upgrade would pass, but it turns out pulpcore 3.11
includes a migration which would handle this for us, if we were
running the db migration before collectstatic

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -5,7 +5,7 @@ class foreman_proxy_content::params {
 
   $qpid_router_sasl_password = extlib::cache_data('foreman_cache_data', 'qpid_router_sasl_password', extlib::random_password(16))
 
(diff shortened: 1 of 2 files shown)


jlsherrill:
corresponds with https://github.com/theforeman/puppet-pulpcore/pull/190   going to run an upgrade pipeline to verify this works as it should

wbclark:
This now appears to be waiting on https://pulp.plan.io/issues/8712

jlsherrill:
This should be good to go now that 3.11.2 has been built and is available in the repos.  If we merge this as is, this HAS to go into katello 4.1.

jlsherrill:
also ping @ekohl for awareness

ekohl (approved):

PR #357, opened 2021-05-08
Fixes #32567: Ensure bootstrap RPM symlink exists

ehelms:
Adds a separate existence check for the symlink to ensure it exists
and points at the latest RPM. This ensures that if something removes
the symlink on the system tne next puppet run will put the symlink
back in place. Prior to this the symlink could be removed from the
system and the only way Puppet would put it back was if the RPM itself
got updated.


ekohl (approved):
In the concat module a resource is generated: https://github.com/puppetlabs/puppetlabs-concat/blob/70070beb18b04cd8bd8fb6c9556848bdd0b5f915/lib/puppet/type/concat_file.rb#L323-L354

I wonder if we could do the same thing here.

I'm fine merging this now. Just good to remember that is also a thing if you need it.

PR #358, opened 2021-05-17
Release 18.2.0

ehelms:
Foreman 2.5 needs the two bug fixes, and would benefit from the enhancements. This would require performing a pulpcore 3.4.0 release (https://github.com/theforeman/puppet-pulpcore/pull/188).

The other option is to split off an 18.1-stable branch and cherry pick the two bug fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,21 @@
 # Changelog
 
-## [18.1.0](https://github.com/theforeman/puppet-foreman_proxy_content/tree/18.1.0) (2021-04-28)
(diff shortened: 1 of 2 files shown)


wbclark (commented):
I'm OK with including the features and doing 18.2.0 , because it seems like https://projects.theforeman.org/issues/32383 is nice to have sooner rather than later

In case we go that route, this would need to require puppet-pulpcore >= 3.4.0

I'll go ahead and update https://github.com/theforeman/puppet-pulpcore/pull/188

ekohl:
I'm ok with pulling this into 2.5. It's still backwards compatible and can actually help with the migration by getting the support in earlier.

PR #369, opened 2021-06-22
Allow puppet-foreman_proxy 19.0.0

ehelms:


ehelms:
This will be needed for puppet-foreman_proxy to drop Puppet 5 and in theory also to pick up this change https://github.com/theforeman/puppet-foreman_proxy/pull/682

wbclark (approved):
I gave an ACK to https://github.com/theforeman/puppet-foreman_proxy/pull/683 so I'll give it here as well

PR #371, opened 2021-06-29
Refs #32891 - Optionally enable new Pulpcore tasking system

wbclark:
Requires https://github.com/theforeman/puppet-pulpcore/pull/203

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -77,6 +77,10 @@
 #                                              degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should be done
 #                                              incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
+# $pulpcore_use_rq_tasking_system::            Use the older RQ workers tasking system instead of the newer PostgreSQL tasking system introduced in Pulpcore 3.13.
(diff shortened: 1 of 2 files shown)


ehelms:
We discussed today that we should enable the new tasking as the default since it's the future and solves a lot of the problems. We didn't discuss at what level we should set the default:

 * pulpcore
 * foreman_proxy_content
 * installer

At a minimum, I think the last two, and for consistency, might as well be all three.

wbclark:
> We discussed today that we should enable the new tasking as the default since it's the future and solves a lot of the problems. We didn't discuss at what level we should set the default:
> 
>     * pulpcore
> 
>     * foreman_proxy_content
> 
>     * installer
> 
> 
> At a minimum, I think the last two, and for consistency, might as well be all three.

Agree it should be the default. Originally when I was pursuing backwards compatibility, it made sense to me to take an incremental approach with default initially off and then change the default over to on.

I think given the timeline, and the agreement that backwards compatibility isn't necessary, then I can go ahead and make the change over in all places.

ehelms:
Since https://github.com/theforeman/puppet-foreman_proxy_content/pull/372 contains the metadata version update, I'll merge this one after it goes in.

ehelms (approved):

PR #372, opened 2021-07-07
Fixes #32910: Support enabling content caching for Pulp

ehelms:
Requires https://github.com/theforeman/puppet-pulpcore/pull/204


ehelms:
I figured exposing these was the cleanest way rather than relying on users configuring hiera.

ekohl (changes requested):
Could you raise the minimum version in metadata.json?

ehelms:
Updated

ekohl (approved):

PR #375, opened 2021-07-12
19.0 cherry picks

ehelms:


manifests/params.pp
```diff
@@ -7,8 +7,9 @@
...
   $pulpcore_allowed_content_checksums = ['sha1', 'sha224', 'sha256', 'sha384', 'sha512']
 
-  $pulpcore_postgresql_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32))
-  $pulpcore_worker_count        = min(8, $facts['processors']['count'])
+  $pulpcore_postgresql_password   = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32))
+  $pulpcore_worker_count          = min(8, $facts['processors']['count'])
+  $pulpcore_use_rq_tasking_system = false
 
   $pulpcore_content_service_worker_timeout = 90
   $pulpcore_api_service_worker_timeout     = 90
```
ekohl, line 15: I'm now looking at these and we should have put these statically in init.pp. Only those that Kafo can't handle should be in params.pp. Given it's a cherry pick, this is OK but we should fix it in master.

ekohl (approved):

manifests/init.pp
```diff
@@ -77,12 +77,21 @@
 #                                              degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should be done
 #                                              incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
+# $pulpcore_use_rq_tasking_system::            Use the older RQ workers tasking system instead of the newer PostgreSQL tasking system introduced in Pulpcore 3.13.
```
wbclark, line 80:
```suggestion
# $pulpcore_use_rq_tasking_system::            Use the older RQ workers tasking system instead of the newer PostgreSQL tasking system introduced in Pulpcore 3.14.
```

I'll open a PR to fix the typo in the main branch as well

PR #378, opened 2021-07-13
Allow to deploy reverse proxy on multiple ports

wbclark:
The reverse proxy will need to listen on both 443 and 8443 to support new clients using 443 by default and migrating existing clients from 8443 to 443

### DIFF
--- a/manifests/reverse_proxy.pp
+++ b/manifests/reverse_proxy.pp
@@ -5,7 +5,7 @@
 # @param url
 #   The URL to forward to
 # @param port
-#   The port to listen on
+#   Port or Array of ports to listen on. One virtual host will be created per port.
 # @param ssl_protocol
 #   The ssl protocol(s) to accept
 # @param vhost_params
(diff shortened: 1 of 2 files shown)


ehelms:
The problem you will quickly hit with this is that we already deploy a virtual host on 443 for Pulp content. I think you will need to take the fragment route similar to https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/init.pp#L217 where instead of shared with the foreman vhost, we now need to share with the Pulp vhost.

templates/apache-fragment.epp
```diff
@@ -0,0 +1,60 @@
+<%- |
```
ekohl, line 1: As noted yesterday, the approach taken in https://github.com/puppetlabs/puppetlabs-apache/pull/1659 feels better to me. I've opened https://github.com/puppetlabs/puppetlabs-apache/pull/2167 which is where I think we should go. Having that could also clean up the puppet-pulpcore module.

manifests/reverse_proxy.pp
```diff
@@ -26,43 +26,57 @@
...
+    Array        => $port,
+    Stdlib::Port => [$port],
+    default      => ['8443'],
+  }
+
+  $ports.each |Stdlib::Port $current_port| {
+    apache::vhost { "${vhost_name}:${current_port}":
+      servername             => $certs::apache::hostname,
+      aliases                => $certs::apache::cname,
+      port                   => $current_port,
```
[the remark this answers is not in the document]
wbclark: I'm planning to attach the pulpcore fragment to this, and want to avoid having pulpcore listen on 8443

with that said, I've been trying a different approach

templates/apache-fragment.epp
wbclark: I've opened https://github.com/ekohl/puppetlabs-apache/pull/1 to get us closer to having a working version

wbclark:
Closing as a much improved version of this was merged already with #390

wbclark:
Closing as a much improved version of this was merged already with #390

PR #383, opened 2021-09-01
Dynamically determine if Puppet should be included

ekohl:
This changes the type to Optional[Boolean]. If it's not explicitly set, some heuristics are used.

It first looks at if the class is already declared, which means it's safe to include it (essentially a noop). This depends on the compilation order.

In order to depend less on the compilation order, it looks up the key `classes` in Hiera. This key is set by Kafo and means it will be included eventually. That means it's safe to include as well.

Users who use this module directly (without Kafo) and use the Hiera `classes` key should set the boolean explicitly.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -20,7 +20,7 @@
 #
 # === Advanced parameters:
 #
-# $puppet::                                    Enable puppet
+# $puppet::                                    Enable puppet. If undef, it will try to detect this automatically.
 #
 # $reverse_proxy::                             Add reverse proxy to the parent
 #
@@ -95,7 +95,7 @@
 class foreman_proxy_content (
   Boolean $pulpcore_mirror = false,
 
-  Boolean $puppet = true,
+  Optional[Boolean] $puppet = undef,
 
   Boolean $reverse_proxy = false,
   Stdlib::Port $reverse_proxy_port = 8443,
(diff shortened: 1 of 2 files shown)


ekohl:
https://github.com/theforeman/foreman-installer/pull/717#discussion_r700210302 perhaps it should be integrated it in puppet-puppet itself.

manifests/init.pp
```diff
@@ -316,7 +316,8 @@
     require               => Class['pulpcore'],
   }
 
-  if $puppet {
+  # classes is set by Kafo
+  if pick($puppet, defined(Class['puppet']) or 'puppet' in lookup('classes', Array, undef, [])) {
```
ehelms, line 320: These types of changes always feel ugly to me -- where we are balancing building a puppet module and wrapping that with kafo and then backing in "kafo" knowledge to a module. Anyway, does not block just stating it as it more tightly links functionality to Kafo design.

ehelms:
This got me wondering what we do if a user is bringing their own puppet infrastructure and needs certificates to be able to talk to Foreman and is using Katello. That question is not specific to this change, but generally speaking. This got me thinking since its the section that generates those certificates but is gated by server being enabled.

ekohl:
Right, that's exactly the problem we need to solve. And solve it in multiple places. I really think we need to get rid of this code in this module but I'm not sure what the best way is now. I'm still wondering what the best design is and struggling with it. You sort of need a data provider interface.

ekohl:
An alternative would be https://github.com/theforeman/puppet-puppet/pull/801. That feels cleaner. It means we align Puppet in all scenarios. We just set the boolean somewhere in Hiera. Please have a look.

ekohl:
I prefer to pursue https://github.com/theforeman/puppet-puppet/pull/801 instead.

PR #384, opened 2021-09-02
Configure RHSM url in Foreman Proxy's Pulp plugin

wbclark:
Requires https://github.com/theforeman/puppet-foreman_proxy/pull/700

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -165,10 +165,6 @@ class foreman_proxy_content (
   include certs::foreman_proxy
   Class['certs::foreman_proxy'] ~> Service['foreman-proxy']
 
-  class { 'foreman_proxy_content::bootstrap_rpm':
-    rhsm_port => $rhsm_port,
-  }
-
   if $reverse_proxy_real {
     class { 'foreman_proxy_content::reverse_proxy':
       url  => "${foreman_url}/",
@@ -328,6 +324,11 @@ class foreman_proxy_content (
     pulpcore_content_url  => "https://${servername}${pulpcore::apache::content_path}",
     client_authentication => ['client_certificate'],
     require               => Class['pulpcore'],
+  }
+
+  class { 'foreman_proxy_content::bootstrap_rpm':
+    rhsm_hostname => $servername,
+    rhsm_port     => $rhsm_port,
   }
 
   if $puppet {
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -155,6 +155,7 @@ class foreman_proxy_content (
   # TODO: doesn't allow deploying a Pulp non-mirror without Foreman
   $shared_with_foreman_vhost = !$pulpcore_mirror
 
+  $rhsm_path = '/rhsm'
   $rhsm_port = $reverse_proxy_real ? {
     true  => $reverse_proxy_port,
     false => 443
@@ -317,18 +318,22 @@ class foreman_proxy_content (
   }
   include pulpcore::plugin::certguard # Required to be present by Katello when syncing a content proxy
 
+  $rhsm_url = "https://${servername}:${rhsm_port}${rhsm_path}"
+
   class { 'foreman_proxy::plugin::pulp':
     pulpcore_enabled      => true,
     pulpcore_mirror       => $pulpcore_mirror,
     pulpcore_api_url      => "https://${servername}",
     pulpcore_content_url  => "https://${servername}${pulpcore::apache::content_path}",
     client_authentication => ['client_certificate'],
+    rhsm_url              => $rhsm_url,
     require               => Class['pulpcore'],
   }
 
   class { 'foreman_proxy_content::bootstrap_rpm':
     rhsm_hostname => $servername,
     rhsm_port     => $rhsm_port,
+    rhsm_path     => $rhsm_path,
   }
 
   if $puppet {
--- a/lib/puppet/provider/rhsm_reconfigure_script/rhsm_reconfigure_script.rb
+++ b/lib/puppet/provider/rhsm_reconfigure_script/rhsm_reconfigure_script.rb
@@ -42,7 +42,7 @@ Puppet::Type.type(:rhsm_reconfigure_script).provide(:rhsm_reconfigure_script) do
       KATELLO_DEFAULT_CA_CERT=<%= resource[:default_ca_name] %>.pem
 
       CERT_DIR=/etc/rhsm/ca
-      PREFIX=/rhsm
+      PREFIX=<%= resource[:rhsm_path] %>
       CFG=/etc/rhsm/rhsm.conf
       CFG_BACKUP=$CFG.kat-backup
       CA_TRUST_ANCHORS=/etc/pki/ca-trust/source/anchors
--- a/lib/puppet/type/rhsm_reconfigure_script.rb
+++ b/lib/puppet/type/rhsm_reconfigure_script.rb
@@ -31,6 +31,11 @@ Puppet::Type.newtype(:rhsm_reconfigure_script) do
     desc "Port that RHSM should be configured to talk to"
   end
 
+  newparam(:rhsm_path) do
+    desc "Path that RHSM should be configured to talk to"
+    defaultto '/rhsm'
+  end
+
   autorequire(:file) do
     [
       self[:server_ca_cert],
--- a/manifests/bootstrap_rpm.pp
+++ b/manifests/bootstrap_rpm.pp
@@ -4,6 +4,7 @@
 class foreman_proxy_content::bootstrap_rpm (
   Stdlib::Fqdn $rhsm_hostname = $facts['networking']['fqdn'],
   Stdlib::Port $rhsm_port = 443,
+  Pattern[/\A(\/[a-zA-Z0-9]+)+(\/)?\z/] $rhsm_path = '/rhsm',
   Stdlib::Absolutepath $rpm_serve_dir = '/var/www/html/pub',
 ) {
   include certs
@@ -54,6 +55,7 @@ class foreman_proxy_content::bootstrap_rpm (
     server_ca_name  => $server_ca_name,
     rhsm_hostname   => $rhsm_hostname,
     rhsm_port       => $rhsm_port,
+    rhsm_path       => $rhsm_path,
   }
 
   bootstrap_rpm { $bootstrap_rpm_name:
--- a/metadata.json
+++ b/metadata.json
@@ -26,7 +26,7 @@
     },
     {
       "name": "theforeman/foreman_proxy",
-      "version_requirement": ">= 19.0.0 < 20.0.0"
+      "version_requirement": ">= 19.1.0 < 20.0.0"
     },
     {
       "name": "katello/qpid",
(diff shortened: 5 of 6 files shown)


lib/puppet/functions/foreman_proxy_content/path_from_url.rb
```diff
@@ -0,0 +1,32 @@
...
+# @example Calling the function
+#   $path = foreman_proxy_content::path_from_url('http://test.example.com')
+#   Returns ''
+#
+Puppet::Functions.create_function(:'foreman_proxy_content::path_from_url') do
+  # @param url URL to parse
+  # @return The path found within the URL
+  dispatch :path do
+    param 'Stdlib::HTTPUrl', :url
+    return_type 'String'
```
wbclark, line 26: Didn't find a suitable type for URL path but I could update this with a `Pattern`.

The regex that `URI` uses frustratingly has comments resembling named capture groups, but not the actual names:

```
irb(main):001:0> URI.regexp
=> 
/
        ([a-zA-Z][\-+.a-zA-Z\d]*):                           (?# 1: scheme)
        (?:
           ((?:[\-_.!~*'()a-zA-Z\d;?:@&=+$,]|%[a-fA-F\d]{2})(?:[\-_.!~*'()a-zA-Z\d;\/?:@&=+$,\[\]]|%[a-fA-F\d]{2})*)                    (?# 2: opaque)
        |
           (?:(?:
             \/\/(?:
                 (?:(?:((?:[\-_.!~*'()a-zA-Z\d;:&=+$,]|%[a-fA-F\d]{2})*)@)?        (?# 3: userinfo)
                   (?:((?:(?:[a-zA-Z0-9\-.]|%\h\h)+|\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}|\[(?:(?:[a-fA-F\d]{1,4}:)*(?:[a-fA-F\d]{1,4}|\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})|(?:(?:[a-fA-F\d]{1,4}:)*[a-fA-F\d]{1,4})?::(?:(?:[a-fA-F\d]{1,4}:)*(?:[a-fA-F\d]{1,4}|\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}))?)\]))(?::(\d*))?))? (?# 4: host, 5: port)
               |
                 ((?:[\-_.!~*'()a-zA-Z\d$,;:@&=+]|%[a-fA-F\d]{2})+)                 (?# 6: registry)
               )
             |
             (?!\/\/))                           (?# XXX: '\/\/' is the mark for hostport)
             (\/(?:[\-_.!~*'()a-zA-Z\d:@&=+$,]|%[a-fA-F\d]{2})*(?:;(?:[\-_.!~*'()a-zA-Z\d:@&=+$,]|%[a-fA-F\d]{2})*)*(?:\/(?:[\-_.!~*'()a-zA-Z\d:@&=+$,]|%[a-fA-F\d]{2})*(?:;(?:[\-_.!~*'()a-zA-Z\d:@&=+$,]|%[a-fA-F\d]{2})*)*)*)?                    (?# 7: path)
           )(?:\?((?:[\-_.!~*'()a-zA-Z\d;\/?:@&=+$,\[\]]|%[a-fA-F\d]{2})*))?                 (?# 8: query)
        )
        (?:\#((?:[\-_.!~*'()a-zA-Z\d;\/?:@&=+$,\[\]]|%[a-fA-F\d]{2})*))?                  (?# 9: fragment)
      /x
irb(main):002:0> URI.regexp.names
=> []
```

manifests/init.pp
```diff
@@ -149,18 +149,30 @@
...
+    Class['certs::apache'] ~> Class['pulpcore::apache']
+    $servername = $certs::apache::hostname
+  }
+
   $rhsm_port = $reverse_proxy_real ? {
     true  => $reverse_proxy_port,
     false => 443
   }
 
+  $rhsm_url = "https://${servername}:${rhsm_port}/rhsm"
```
ehelms, line 167: What is the advantage of constructing this here, and then deconstructing it again here (https://github.com/theforeman/puppet-foreman_proxy_content/pull/384/files#diff-0393165ecfe8896424ac06417b31d37ab529a477ea8982c4573945a31a28e43aR20-R22)? 

I am looking to understand the benefit of adding the complexity of deconstruction when we were already passing the individual parameters to the class. That seems to add code into the process that could be avoided by passing the parts in directly as we already do today.

I would add on to that, that we plan to drop the boostrap RPM in a few coming releases, so thats a factor in just how much support code we add in this area.

wbclark: My thinking has mostly been around treating `$rhsm_url` as a single parameter and eventually single source of truth for the `rhsm_hostname`, `rhsm_port`, and `rhsm_path`.

I considered (and previously wrote it as) changing the interface to the bootstrap RPM type to accept the `$rhsm_url` and use `URI::Parse` inside the script template to perform the deconstruction there. I explored this approach instead to avoid the breaking change to that type interface, moving the deconstruction outside of the provider.

But the way I currently wrote it does instead cause a breaking change in the `foreman_proxy_content::bootstrap_rpm` class so I agree it's kind of pointless. I am tempted to preserve that class interface as well and perform the deconstruction inside `init.pp` in order to preserve interfaces while ensuring that there is a single source of truth for `$rhsm_url` and its constituent pieces that will be propagated to the bootstrap RPM and smart_proxy_pulp plugin.

Longer term I had the idea of exposing some parameters in the `foreman_proxy_content` class, that may default to `undef` but if defined would take precedence over `$servername`, `$reverse_proxy_real ? { true => $reverse_proxy_port, false => 443 }`, and `/rhsm` (perhaps there is little use for that last one though) in the way the `$rhsm_url` gets constructed. I see this as being more forward thinking towards the possibility of HA services and a more modular design for the SmartProxy. (It would also require separating the rhsm_url out of the smartproxy_pulp plugin and into @ekohl 's idea for smartproxy_rhsm plugin)

wbclark: I also want to state that I have some skepticism about the removal of the bootstrap RPM, based on separation in some deployments between katello users (meaning someone who know what katello is, has a login) vs. their end users (maybe someone in a different team or department, or sometimes even an external customer) who may be a sysadmin that doesn't know about or have any access to touch Katello at all, and only knows that in order to get subscriptions and repositories they do: `rpm -Uvh http://smartproxy.example.com/pub/katello-ca-consumer-latest.noarch.rpm && subscription-manager register --activationkey=$key1,$key2,$key3`

For that reason, my preference would be not to remove the functionality of the bootstrap RPM entirely. I understand that global registration is a better process for a heavily automated, cloud first world, but many users, and particularly many of the _user's users_, are not there yet and might prefer not to deprecate the current registration workflow. I think I recall that there are some other concerns with the bootstrap RPM as well. Perhaps keeping it around in an opt-in fashion (i.e. `Boolean generate_and_deploy_bootstrap_rpm = false` in the interface for the main `foreman_proxy_content` class) is still better to serve a wider swath of users?

manifests/init.pp
```diff
@@ -149,18 +149,30 @@
...
   $shared_with_foreman_vhost = !$pulpcore_mirror
 
+  if $shared_with_foreman_vhost {
+    include foreman::config::apache
+    Class['foreman::config::apache'] -> Class['pulpcore::apache']
+    $servername = $foreman::config::apache::servername
+  } else {
+    include certs::apache
+    Class['certs::apache'] ~> Class['pulpcore::apache']
+    $servername = $certs::apache::hostname
```
ehelms, line 159: I think this is true in all cases and could be pulled out of the if-else. I realize we did not do this before but we could have.

ehelms:
I'd be interested to see a scaled back version of this. That is, what would the minimally required changes be to pass the `rhsm_url` be given the information already preset and the interface of the classes staying in tact and being added to? You could then do a completely separate PR that looks at changing the interface and introducing additional functions.

manifests/init.pp
ehelms: I think if there are concerns over moving to global registration as the one and only method for how we do this procedure we should discuss that more broadly. Looking through our RFCs I do not see one dedicated to the discussion of deprecating these various workflows. I will look at opening one and then you can elevate your thoughts there for broader discussion.

ehelms: Here is said RFC -- https://community.theforeman.org/t/rfc-deprecating-katello-bootstrap-client-tools/25220

manifests/init.pp
wbclark: Which part specifically -- the include and the ordering relationship? Shouldn't that be causing a problem currently if it's required to be defined outside of the else?

Pulling the `$servername = ...` out of the else here would cause a problem, because it is defined differently under the if.

wbclark: I did find that we `include certs::apache` in manifests/reverse_proxy.pp; and `foreman_proxy_content::reverse_proxy` will be used when `$pulpcore_mirror or $reverse_proxy`, meanwhile we do explicitly include it here when `$shared_with_foreman_vhost` (i.e. when `!$pulpcore_mirror`).

So since it gets ultimately included when `$pulpcore_mirror or !$pulpcore_mirror`, that is always. I think I see your point that it could be included at the top level here. And that passes the sanity test -- when do we not need apache and its certs?

But what if the conditions change in the future? For example if we no longer have `$shared_with_foreman_vhost = !$pulpcore_mirror` but instead some other logic? I think I prefer to let puppet handle that thinking for us instead of opening up the possibility of always including something we may no longer need in the future?

ehelms: I am most saying, that for the forseeable future, and today, the servername where we deploy Apache is always mapped to `    $servername = $certs::apache::hostname` and in some ways is the best source of truth for this value given these are the certificates under which Apache is running.

ekohl: > I am most saying, that for the forseeable future, and today, the servername where we deploy Apache is always mapped to ` $servername = $certs::apache::hostname` and in some ways is the best source of truth for this value given these are the certificates under which Apache is running.

This is also what I always used as leading. While technically there are `cnames` and `subjectAltName`, we can also assume a `commonName` in our certificates and assume that works.

Technically you could also do it the other way around and have a `$servername` parameter and call `class { 'certs::apache': hostname => $servername }` but that's effectively the same thing.

ekohl (changes requested):
I took a stab at it in https://github.com/theforeman/puppet-foreman_proxy_content/pull/385. My goal there was to keep it small while still achieving the same thing. Hopefully it shows that it's easier to review the changes since less things move around.

wbclark:
@ekohl @ehelms 

I plan this morning to update this with changes to the vhost fragment template that move some configurations outside of the `<Location>` context as necessary, and ensure other features of the existing 8443 reverse proxy such as configurable timeouts are present.

At this point, would it be better to split the commit "Configure pulpcore mirror with reverse proxy on 443 for rhsm_path" into a separate PR to continue that work, so that the others can be reviewed and merged, or is it better to leave this as a single PR?

ehelms:
> At this point, would it be better to split the commit "Configure pulpcore mirror with reverse proxy on 443 for rhsm_path" into a separate PR to continue that work, so that the others can be reviewed and merged, or is it better to leave this as a single PR?

I think if you want to split some of that out into a separate PR to get in ahead of time and reduce the size of this PR that works for me as some of the changes are stand-alone and mergable.

wbclark:
I removed the commit here so that this PR remains focused only on configuring the RHSM URL.

Then I started a new approach of changing the RHSM port to 443 which is in the PR #390

ekohl (approved):

PR #387, opened 2021-10-20
Fixes #33751 - enable python and ostree content

jlsherrill:


jlsherrill:
yep, added here: https://github.com/theforeman/puppet-pulpcore/pull/236

will add that test!

PR #393, opened 2021-12-16
Fixes #34154 - Refresh on Pulp plugin installation

ekohl:
smart_proxy_pulp dynamically retrieves the Pulp content types and Katello uses this. This means the features need to be refreshed after a content type is added. The API also changes so that cache needs to be regenerated.

This does refreshes on the foreman_smartproxy which could be considered a private API from theforeman/foreman_proxy and collectors would silently be skipped if anything changes. That's why this must be covered by spec tests.

Currently untested on an actual system (which is why this is now a draft) but locally the unit tests passed and having a PR makes it easier to patch a forklift VM.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -352,6 +352,17 @@ class foreman_proxy_content (
     rhsm_path     => $rhsm_path,
   }
 
+  # smart_proxy_pulp dynamically retrieves the Pulp content types and Katello
+  # uses this. This means the features need to be refreshed after a content
+  # type is added. The API also changes so apipie cache needs to be regenerated.
+  # lint:ignore:spaceship_operator_without_tag
+  if $foreman_proxy::register_in_foreman {
+    Pulpcore::Plugin <| |> ~> Foreman_smartproxy[$foreman_proxy::registered_name]
+    Foreman_smartproxy[$foreman_proxy::registered_name] -> Foreman::Rake <| title == 'apipie:cache:index' |>
+  }
+  Pulpcore::Plugin <| |> ~> Foreman::Rake <| title == 'apipie:cache:index' |>
+  # lint:endignore
+
   if $puppet {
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -352,6 +352,17 @@
...
 
+  # smart_proxy_pulp dynamically retrieves the Pulp content types and Katello
+  # uses this. This means the features need to be refreshed after a content
+  # type is added. The API also changes so apipie cache needs to be regenerated.
+  # lint:ignore:spaceship_operator_without_tag
+  if $foreman_proxy::register_in_foreman {
+    Pulpcore::Plugin <| |> ~> Foreman_smartproxy[$foreman_proxy::registered_name]
+    Foreman_smartproxy[$foreman_proxy::registered_name] -> Foreman::Rake <| title == 'apipie:cache:index' |>
+  }
+  Pulpcore::Plugin <| |> ~> Foreman::Rake <| title == 'apipie:cache:index' |>
```
ekohl, line 363: Testing shows that `foreman-rake apipie:cache:index` doesn't help. This needs the full blown `foreman-rake apipie:cache` which we don't do, only in packaging. That's because it's expensive and SLOW. I'm not sure how to deal with that.

ekohl: Note that this doesn't need to be a blocker: the `apipie:cache:index` at least generates the correct JSON which FAM and Hammer can use.

ehelms: Does that imply we should drop this line for now?

evgeni: My reading is: no, it should remain, this patch just doesn't fix "all" of the problems in the redmine.

But we can probably just split that.

ekohl: @evgeni is right: the index regeneration is certainly needed (fixes FAM and Hammer) but it's not all since the human readable HTML isn't regenerated.

evgeni (approved):

PR #395, opened 2022-01-27
Refs #34325 - enable redis cache by default

jlsherrill:


ekohl (approved):

ekohl:
You can ignore the Puppet 6 unit test failures. It's a CI issue. We'll see about the other tests.

evgeni:
EL8 fails due to the new module repodata

I propose #396 as a fix, or properly using `katello::repository` ;)

ekohl:
I'd be ok with putting the katello module in fixtures and reusing that.

Slight side conversation: long term I'd actually think we should push more from puppet-katello into puppet-foreman. The repo definition could be `foreman::repo::katello` and the actual plugin could be `foreman::plugin::katello`. Since this module already  pulls in the foreman module, it would be solved.

PR #396, opened 2022-01-28
use katello::repo to setup katello repo for tests 

evgeni:


evgeni:
```
       	Error: Execution of '/usr/bin/dnf -d 0 -e 1 -y install /var/www/html/pub/katello-ca-consumer--1.0-1.noarch.rpm' returned 1: Could not open: /var/www/html/pub/katello-ca-consumer--1.0-1.noarch.rpm
```

I think something is broken with the hostname detection, als noted by @ekohl yesterday evening.

ekohl:
I looked at including `katello::repo` in puppet-foreman but ran into the problem that the version format is different. Tomorrow I'll take a look. For now I just restarted the tests to see if it's solved now with the new beaker-hostgenerator.

evgeni:
Looks as :green_apple: as it gets :fireworks:

PR #399, opened 2022-02-04
Require at least certs 15

ekohl:
certs used to have this, but in version 14 it was absent. Raising the minimum version to 15 is the easiest way to guarantee it is really there.

Fixes: a3ac43579b1f456b11c989a3ef0230718a36a44e

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -34,7 +34,7 @@
     },
     {
       "name": "katello/certs",
-      "version_requirement": ">= 11.0.0 < 16.0.0"
+      "version_requirement": ">= 15.0.0 < 16.0.0"
     },
     {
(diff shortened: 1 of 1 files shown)


evgeni (approved):
Is this fixing an actual bug, or just playing safe?

(I'm fine with either, just curious)

ekohl:
https://github.com/theforeman/puppet-certs/commit/16478ba977522e79cf0fa178bdf811d199acb39e reintroduced the variable that https://github.com/theforeman/puppet-certs/commit/8c5404ff76aca156be16f81b1f21b630a067b2c2 removed. So technically we could have something like `>= 11, < 14 || >= 15` but puppet-katello already requires `>= 15` anyway.

PR #400, opened 2022-02-07
Fixes #34406 - drop support for ancient EL versions

evgeni:
we don't need to support EL5 at all anymore, and EL6 has the fixed
sub-man sice 6.2, which is totally good enough


ekohl:
How breaking is this? I'm looking at the labels Bug, Enhancement and Backwards incompatible and somehow feel like the latter might be correct but also wonder if you want to cherry pick this.

evgeni:
In Foreman, we dropped support for EL5 and EL6 clients long ago (6 in 3.0, 5 I don't even remember), so "not breaking" unless we pick it further than 3.0, which I don't intend to.

PR #405, opened 2022-03-30
Refs #34684 - enable pulp-cli for Katello

evgeni:


.fixtures.yml
```diff
@@ -20,7 +20,9 @@ fixtures:
     xinetd:        "https://github.com/puppetlabs/puppetlabs-xinetd"
     certs:         "https://github.com/theforeman/puppet-certs.git"
     qpid:          "https://github.com/theforeman/puppet-qpid.git"
-    pulpcore:      "https://github.com/theforeman/puppet-pulpcore.git"
+    pulpcore:
+      repo: "https://github.com/evgeni/puppet-pulpcore.git"
+      branch: "pulp-cli"
```
evgeni, line 25: this is just here to run tests before https://github.com/theforeman/puppet-pulpcore/pull/252 is merged ;)

manifests/init.pp
```diff
@@ -287,6 +287,16 @@
     before                         => Class['foreman_proxy::plugin::pulp'],
   }
 
+  if ! $pulpcore_mirror {
```
evgeni, line 290: the mirrors (aka proxies) don't have the certs required

ekohl: I think this is technically a bit more accurate?
```suggestion
  if $shared_with_foreman_vhost {
```

Technically it's the same right now, but the idea behind it is that it could end up deploying just a content proxy without Foreman but still is a non-mirror. Essentially you could end up with `foreman.example.com` (with Katello, without Pulp) and `pulp.example.com` (primary content storage). Then you could also have `mirror.location.example.com` which would be a local mirror.

evgeni: Possible, yeah. I must admit I did not exactly understand the difference from the code. Your explanation makes sense :)

ekohl: I captured my thoughts in https://github.com/theforeman/puppet-foreman_proxy_content/issues/407

ekohl:
This inspired me to finish https://github.com/theforeman/puppet-foreman_proxy_content/pull/326 but as always you end up writing other patches so that's waiting for https://github.com/theforeman/puppet-foreman_proxy/pull/733. If we have that we should be able to verify it end-to-end.

ehelms (approved):

ekohl (approved):

PR #411, opened 2022-04-07
Pre-install pulpcore-selinux in acceptance tests

wbclark:
Otherwise the test will fail idempotence in an selinux enabled
environment due to Puppet deciding selinux labels at the time of
catalogue compilation.

### DIFF
--- a/spec/setup_acceptance_node.pp
+++ b/spec/setup_acceptance_node.pp
@@ -8,8 +8,11 @@ class { 'katello::repo':
 
 include pulpcore::repo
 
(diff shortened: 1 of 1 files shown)


wbclark:
This was an issue I ran into while running the new content acceptance tests on my machine

evgeni:
I'd be also OK with this if Ewouds inline comment is addressed.

But can Puppet finally behave properly when policies are installed during the run?! It has bitten us several times and I bet more is to come.

evgeni (approved):

PR #412, opened 2022-04-20
Bump theforeman/pulpcore requirement to >= 6.0.0

evgeni:
We need 6.0.0 for the `pulpcore::cli` support

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -38,7 +38,7 @@
     },
     {
       "name": "theforeman/pulpcore",
-      "version_requirement": ">= 4.0.0 < 6.0.0"
+      "version_requirement": ">= 6.0.0 < 7.0.0"
     }
   ],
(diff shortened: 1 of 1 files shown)


evgeni:
This won't pass CI before https://github.com/theforeman/puppet-pulpcore/pull/255 is merged, so I'll just do that ;)

ehelms:
> ~This won't pass CI before [theforeman/puppet-pulpcore#255](https://github.com/theforeman/puppet-pulpcore/pull/255) is merged, so I'll just do that ;)~
> 
> Heh, it did pass…

Tests use https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/.fixtures.yml#L23 but this would break when the installer nightly job ran trying to reconcile module dependencies.

ehelms (approved):

PR #416, opened 2022-04-28
Release 21.0.0

evgeni:


CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
 # Changelog
 
+## [21.0.0](https://github.com/theforeman/puppet-foreman_proxy_content/tree/21.0.0) (2022-04-28)
+
+[Full Changelog](https://github.com/theforeman/puppet-foreman_proxy_content/compare/20.2.0...21.0.0)
+
+**Breaking changes:**
+
+- Refs [\#34640](https://projects.theforeman.org/issues/34640) - Drop apipie cache work around [\#414](https://github.com/theforeman/puppet-foreman_proxy_content/pull/414) ([ehelms](https://github.com/ehelms))
```
evgeni, line 9: Couldn't really decide whether this is breaking or not, but given we already had one breaking change to release, I went with breaking here too.

CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
...
+- Drop pulpcore\_use\_rq\_tasking\_system parameter [\#404](https://github.com/theforeman/puppet-foreman_proxy_content/pull/404) ([ekohl](https://github.com/ekohl))
+
+**Implemented enhancements:**
+
+- Allow puppet-foreman\_proxy 22.0.0 [\#413](https://github.com/theforeman/puppet-foreman_proxy_content/pull/413) ([ehelms](https://github.com/ehelms))
+- Bump theforeman/pulpcore requirement to \>= 6.0.0 [\#412](https://github.com/theforeman/puppet-foreman_proxy_content/pull/412) ([evgeni](https://github.com/evgeni))
```
evgeni, line 15: wasn't sure about this either -- technically it's "part" of the `pulp-cli` change below, even if it went in in an own PR?

wbclark: Is this not a breaking change even if the API does not actually change? Previously this module allowed you to use older puppet-pulpcore version, now it requires at least 6.0.0. IMO an enhancement would be _allowing_ newer puppet-pulpcore, not _requiring_ it.

evgeni: it very well might be, yes :)

ehelms (approved):

PR #417, opened 2022-05-09
Fixes #34882 - allow custom pulpcore import/export paths

wbclark:


manifests/init.pp
```diff
@@ -92,6 +92,9 @@
 # $pulpcore_cache_expires_ttl::                The number of seconds that content should be cached for.
 #                                              Specify 'None' to never expire the cache.
 #
+# $additional_pulpcore_import_paths::          Additional allowed paths that Pulpcore can use for content imports, or sync from using file:// protocol
+# $additional_pulpcore_export_paths::          Additional allowed paths that Pulpcore can use for content exports
```
ehelms, line 96:
```suggestion
#
# $additional_pulpcore_export_paths::          Additional allowed paths that Pulpcore can use for content exports
```

manifests/init.pp
```diff
@@ -139,6 +142,8 @@
   Integer[0] $pulpcore_api_service_worker_timeout = 90,
   Boolean $pulpcore_cache_enabled = true,
   Optional[Variant[Integer[1], Enum['None']]] $pulpcore_cache_expires_ttl = undef,
+  Variant[Stdlib::Absolutepath, Array[Stdlib::Absolutepath]] $additional_pulpcore_import_paths = [],
+  Variant[Stdlib::Absolutepath, Array[Stdlib::Absolutepath]] $additional_pulpcore_export_paths = [],
```
ehelms, line 146: Since both of these are *plural* I would expect it to only allow array.

manifests/init.pp
ehelms: Perhaps this should also follow the theme of prefix? e.g.

```
pulpcore_additional_expore_paths
pulpcore_additional_import_paths
```

manifests/init.pp
```diff
@@ -215,17 +220,20 @@
   }
 
   if $pulpcore_mirror {
-    $pulpcore_allowed_import_path    = ['/var/lib/pulp/sync_imports']
-    $pulpcore_allowed_export_path    = []
+    $base_allowed_import_path    = ['/var/lib/pulp/sync_imports']
+    $base_allowed_export_path    = []
```
ehelms, line 224: Might as well makes these plural as well since they are arrays and you are introducing a new name.

manifests/init.pp
evgeni: While this is true, passing arrays in the installer is cumbersome and allowing strings here (and convert them later) makes the UX better, IMHO

wbclark: > While this is true, passing arrays in the installer is cumbersome and allowing strings here (and convert them later) makes the UX better, IMHO

This was my reasoning. I kept the option to pass a string for now, let me know if there is other feedback on it.

manifests/init.pp
wbclark: I missed this in the first round of updates that I pushed. But it's a good point, I'll update it again in a moment.

spec/classes/foreman_proxy_content_spec.rb
```diff
@@ -25,6 +27,32 @@
             .with_rhsm_url("https://#{facts[:fqdn]}:443/rhsm")
         end
 
+        context 'with custom import/export paths' do
+          let(:params) do
+            # an array is not used for the export paths, to test that a Stdlib::Absolutepath value
+            # is properly handled as if it were a single element array
```
ehelms, line 33: This has the potential to get lost as it's not as "enforced" as a dedicated test case. I would rather lean towards just adding another context block that is dedicated to testing passing a string to both parameters, e.g.

```
context 'with custom import/export paths as strings' do

end
```

spec/classes/foreman_proxy_content_spec.rb
```diff
@@ -25,6 +27,32 @@
...
+                  '/var/lib/pulp/sync_imports',
+                  '/var/lib/pulp/imports',
+                  '/my/custom/import/path'
+                ]
+              )
+              .with(allowed_export_path: ['/var/lib/pulp/exports', '/my/custom/export/path'])
+          end
+        end
+
+
```
ehelms, line 55:
```suggestion
```

spec/classes/foreman_proxy_content_spec.rb
wbclark: OK, I've created separate test cases for array and string inputs.

ehelms (approved):
@evgeni your turn to look

evgeni (approved):
> @evgeni your turn to look

looks like a puppet manifest

PR #420, opened 2022-06-14
Refs #35005 - use foreman client certs for pulp-cli auth

evgeni:


manifests/reverse_proxy.pp
```diff
@@ -37,7 +37,7 @@
   apache::vhost { $vhost_name:
     ensure                 => $ensure,
     servername             => $certs::apache::hostname,
-    aliases                => $certs::apache::cname,
+    serveraliases          => $certs::apache::cname,
```
evgeni, line 40: this was a bug, should it be an own PR (which would fail tests, but oh well)?

.sync.yml
```diff
@@ -10,5 +10,8 @@ spec/spec_helper.rb:
     - name: root_home
       value: /root
       source: puppetlabs-stdlib
+    - name: service_provider
```
ekohl, line 13: It was my intention to move to `add_mocked_facts!` (from [voxpupuli-test](https://github.com/voxpupuli/voxpupuli-test/blob/master/lib/voxpupuli/test/facts.rb)) but haven't gotten around to it. We should implement this in our modulesynced `spec_helper.rb`. It would make all these custom facts redundant.

manifests/init.pp
```diff
@@ -298,12 +298,11 @@
...
 
   if $shared_with_foreman_vhost {
-    include certs::pulp_client
     class { 'pulpcore::cli':
       pulpcore_url => "https://${servername}",
       dry_run      => true,
-      cert         => $certs::pulp_client::client_cert,
-      key          => $certs::pulp_client::client_key,
+      cert         => $certs::foreman_proxy::foreman_ssl_cert,
+      key          => $certs::foreman_proxy::foreman_ssl_key,
```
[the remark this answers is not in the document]
ehelms: The replacement should be certs::foreman::client_cert

ehelms:
Three commits and one PR is fine

manifests/init.pp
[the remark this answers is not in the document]
evgeni:
```
  No such file or directory @ rb_sysopen - /etc/foreman/client_key.pem
  Error: /Stage[main]/Certs::Foreman/Certs::Keypair[centos8-64.example.com-foreman-client]/Private_key[/etc/foreman/client_key.pem]/ensure: change from 'absent' to 'present' failed: Could not set 'present' on ensure: No such file or directory @ rb_sysopen - /etc/foreman/client_key.pem (file: /etc/puppetlabs/code/environments/production/modules/certs/manifests/keypair.pp, line: 19)
```

:confused:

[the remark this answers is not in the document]
evgeni: Right, one needs `/etc/foreman` present for this to work. Nice.

evgeni:
@ehelms this can now get a nice green checkmark ;-)

ehelms (approved):

manifests/reverse_proxy.pp
ekohl: Now I've cherry picked this, but I don't have a changelog entry. So in hindsight, it would have been nice.

evgeni: well… ;)

ekohl: In https://github.com/theforeman/puppet-foreman_proxy_content/pull/430 I tried to describe it.

PR #422, opened 2022-07-15
Fixes #35227 - Load apache::mod::autoindex if needed

ekohl:
On content proxies we have (by default) /pub as a browsable directory, though it can be disabled by passing a different pub dir option. If it is browsable, mod_autoindex must be loaded. The installer disables loading default mods so it should be loaded explicitly.


evgeni (approved):

ekohl:
In the thread it was mentioned that the `/pub` prefix is somehow missing. Perhaps another module is required for that.

PR #423, opened 2022-07-26
Fixes #35278: Add Red Hat Insights path to reverse proxy URLs

ehelms:


ehelms:
@sshtein are there any other paths, to your knowledge, that need to be reverse proxied back to Foreman through an smart-proxy with content?

manifests/init.pp
```diff
@@ -160,17 +160,19 @@
...
   include certs::foreman_proxy
   Class['certs::foreman_proxy'] ~> Service['foreman-proxy']
 
   if $reverse_proxy_real {
     foreman_proxy_content::reverse_proxy { "rhsm-pulpcore-https-${reverse_proxy_port}":
-      url      => "${foreman_url}/",
-      port     => $reverse_proxy_port,
-      priority => '10',
-      before   => Class['pulpcore::apache'],
+      path_url_map => { '/' => "${foreman_url}${rhsm_path}" },
```
ekohl, line 172: This looks like a change where `/` used to point at `${foreman_url}`, it now points at `${foreman_url}${rhsm_path}`.

ShimShtein:
I think we should already be prepared for the LEAPP stuff too: https://github.com/theforeman/foreman_rh_cloud/pull/750 will add `/api/pes` to the list.

ekohl (commented):
I wonder how this deals with authentication. Does insights use client certificates? If so, how do we (securely) transfer that authentication to Foreman?

ehelms:
> I wonder how this deals with authentication. Does insights use client certificates? If so, how do we (securely) transfer that authentication to Foreman?

The RHSM certificates are used, and authentication is handled in the same way that subscription-manager requests are handled from the machine.

ekohl (approved):

PR #428, opened 2022-09-12
Fixes #35486 - Restart yggdrasil in rhsm-reconfigure

adamruzicka:


lib/puppet/provider/rhsm_reconfigure_script/rhsm_reconfigure_script.rb
```diff
@@ -130,6 +130,9 @@ def script_template
         update-ca-trust
       fi
 
+      # restart yggdrasild if it is installed and running
+      systemctl try-restart yggdrasil >/dev/null 2>&1
```
ekohl, line 134: Something I recently learned is that RHEL 9 kickstart will actually execute this script if you pass `rhsm --server-hostname`. I'm wondering what `systemctl` does in that context. Perhaps it should have `|| true` (since `set -e` is used).

ekohl (approved):

PR #431, opened 2022-10-19
Fixes #35607 - Pulpcore Telemetry should be disabled initially

wbclark:
Introduces a pulpcore_telemetry parameter which is false by default. Users can opt-in to reporting anonymized telemetry data to the pulp project at https://analytics.pulpproject.org/ by setting the value to true.

Requires https://github.com/theforeman/puppet-pulpcore/pull/267

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -96,6 +96,8 @@
 #
 # $pulpcore_additional_export_paths::          Additional allowed paths that Pulpcore can use for content exports
 #
+# $pulpcore_telemetry::                        Enable upload of anonymous usage data to https://analytics.pulpproject.org/
+#
 class foreman_proxy_content (
   Boolean $pulpcore_mirror = false,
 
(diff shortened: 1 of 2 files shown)


sjha4:
Can we re-run the tests? I am not sure the failure is related..I saw another PR with green tests after this PR..

ekohl:
@wbclark could you rebase?

ekohl (changes requested):
The puppet-pulpcore PR should have used `Refs #35607` and this PR should have used `Fixes #35607`. The reason is that puppet-pulpcore was a prerequisite to completing the feature, but didn't actually complete it. Now it's marked as completed which means it no longer shows up in my TODOs.

I've now opened the issue again and set a target version. Could you change this PR to use `Fixes`?

wbclark:
> The puppet-pulpcore PR should have used `Refs #35607` and this PR should have used `Fixes #35607`. The reason is that puppet-pulpcore was a prerequisite to completing the feature, but didn't actually complete it. Now it's marked as completed which means it no longer shows up in my TODOs.

Thank you. I've rebased, changed the commit message, pushed, and updated the PR title.

wbclark:
The error looks different now. Looks like my ruby brain used a lowercase `false` when python only recognizes `False`

ekohl:
Could you also change the pulpcore requirement in metadata.json to at least 7.0.0 (or 7.0.1 when we release a bugfix)?

ekohl (approved):
Thanks! Merging, though technically it needs a release.

PR #432, opened 2022-10-27
Fixes #35681: Use registration_url for RHSM and Pulp content URL 

ehelms:


evgeni:
*url* to me implies that this is the *whole* thing, yet you still append port and path to them which feels counter-intuitive.

I think we either should take the url verbatim and ignore the other parameters in that case, or call it "host" or "base" or whatever

ekohl (changes requested):
I agree. It should also match the `servername` since that's what Apache thinks the vhost is named. So instead I think you want a parameter to override `servername` somehow. Note we also set up the pulpcore CLI to connect to `https://${servername}`.

One way is to expose `certs::apache::hostname` somehow.

Additionally: you may also find out you want `CNAME` support with `serveraliases` on the Apache vhost. We don't have that today.

ehelms:
> . It should also match the `servername` since that's what Apache thinks the vhost is named. So instead I think you want a parameter to override `servername` somehow.

I have updated to hostname instead of URL. I need some further details to better understand the implications of this statement.

ekohl (commented):
My concern with this is that the specified hostname may not be the same as on the certificate, which would lead to verification errors by clients. What I was referring to is in https://github.com/theforeman/puppet-pulpcore/pull/269.

ehelms:
> My concern with this is that the specified hostname may not be the same as on the certificate, which would lead to verification errors by clients. What I was referring to is in [theforeman/puppet-pulpcore#269](https://github.com/theforeman/puppet-pulpcore/pull/269).

If a user is providing custom certificates, then we still need these parameters for them to customize and indicate the desired state. I would agree your proposals make life easier for those that only use the default certificates, but do not solve all use cases. 

To help users end up in the right state for all use cases, I am hearing a need for an installer hook that does some matching verification perhaps? Unless you think that checking can be easily done at the puppet module level.

ekohl:
> If a user is providing custom certificates, then we still need these parameters for them to customize and indicate the desired state. I would agree your proposals make life easier for those that only use the default certificates, but do not solve all use cases.

Yes, it is needed to set the parameters, but I don't see how that's different from your proposal. One difference is that your proposal also allows users to change the port but I'm not sure we should support that. It's already complex enough right now.

ehelms:
> One difference is that your proposal also allows users to change the port but I'm not sure we should support that. It's already complex enough right now.

My updated version is only setting hostname which does not support setting the port.

manifests/init.pp
```diff
@@ -353,20 +359,30 @@
   }
   include pulpcore::plugin::certguard # Required to be present by Katello when syncing a content proxy
 
-  $rhsm_url = "https://${servername}:${rhsm_port}${rhsm_path}"
+  if $rhsm_hostname {
+    $rhsm_url = "https://${rhsm_hostname}:${rhsm_port}${rhsm_path}"
+  } else {
+    $rhsm_url = "https://${servername}:${rhsm_port}${rhsm_path}"
+  }
```
evgeni, line 366:
```suggestion
  $rhsm_hostname_real = pick($rhsm_hostname, $servername)
  $rhsm_url = "https://${rhsm_hostname_real}:${rhsm_port}${rhsm_path}"
```

Same below?

ekohl:
> My updated version is only setting hostname which does not support setting the port.

Then I wonder what's the point. With this solution you're forced to use custom certificates, right? My solution works for both custom certificates and built in certificates. I'm leaning to :-1: on this.

ekohl:
I'll elaborate a bit: you don't set the `servername` to `pulpcore` which means it's not configuring the correct `CONTENT_HOST` and `CONTENT_ORIGIN` settings (and `ANSIBLE_API_HOSTNAME` and `ANSIBLE_CONTENT_HOSTAME` if the ansible plugin is enabled). That means Pulp can't generate the correct URLs.

ehelms:
> I'll elaborate a bit: you don't set the `servername` to `pulpcore` which means it's not configuring the correct `CONTENT_HOST` and `CONTENT_ORIGIN` settings (and `ANSIBLE_API_HOSTNAME` and `ANSIBLE_CONTENT_HOSTAME` if the ansible plugin is enabled). That means Pulp can't generate the correct URLs.

I am going to step back to what my goal is, as the design discussions were what drove us to arrive at this solution.

I want to solve the existing use case that is setting the parameters we need such that client's rhsm.conf get configured with a load balancer as the `hostname =` and as the hostname in the `baseurl=` when they are registering through the registration API.

As I understand it, that configuration is driven off of the `content_app_url` and `rhsm_url` that is reported by the `smart_proxy_pulp` plugin settings.

I am not aware of the need to solve the additional layers that you have identified at this time -- that is not to say we should not explore what those other items should be configured as in a load balanced setup but to my knowledge they are not required to solve the RHSM use case which is an established use case.


(Side note which hopefully does not derail: I cannot find CONTENT_HOST as a setting anymore?)

ekohl:
> I want to solve the existing use case that is setting the parameters we need such that client's rhsm.conf get configured with a load balancer as the `hostname =` and as the hostname in the `baseurl=` when they are registering through the registration API.

Right, so also taking a step back: today we configure all services on the content proxy based on the `certs::apachehostname`. So first we determine the servername:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/63f6732235286b73a76b2452f06b4f2d57bb4836/manifests/init.pp#L259

This is then used in various places. Pulpcore is configured to use it.
https://github.com/theforeman/puppet-foreman_proxy_content/blob/63f6732235286b73a76b2452f06b4f2d57bb4836/manifests/init.pp#L284

That in turn uses it to make Pulpcore aware of how it's reachable externally:
https://github.com/theforeman/puppet-pulpcore/blob/654d7fe22015f0e0ed5e26cb5cc973f54978a480/templates/settings.py.erb#L20-L21
https://github.com/theforeman/puppet-pulpcore/blob/5858c93e91a29f910aa9eab0e0fb6bd34a69e071/manifests/plugin/ansible.pp#L18-L31
It also drives how Pulpcore CLI is configured to reach the server:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/63f6732235286b73a76b2452f06b4f2d57bb4836/manifests/init.pp#L311

Additionally, we configure the RHSM proxy with the correct hostname:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/63f6732235286b73a76b2452f06b4f2d57bb4836/manifests/init.pp#L173-L196

It is really defined here:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/63f6732235286b73a76b2452f06b4f2d57bb4836/manifests/reverse_proxy.pp#L45-L46

Now that the services are configured, it's then exposed here:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/63f6732235286b73a76b2452f06b4f2d57bb4836/manifests/init.pp#L365-L368

And the bootstrap RPM:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/63f6732235286b73a76b2452f06b4f2d57bb4836/manifests/init.pp#L372-L376

So my approach was to configure `certs::apache` to contain the correct values. First of all, this means a correct certificate is generated when using the default certs (custom certs are on their own). Secondly, because all services are configured using these variables we configure the services correctly.

I fail to see how your PR solves the issue in an elegant way. It makes it easy to create mismatches in service configuration and makes it more complex to support.

> (Side note which hopefully does not derail: I cannot find CONTENT_HOST as a setting anymore?)

https://github.com/theforeman/puppet-pulpcore/search?q=servername

evgeni:
certs::cname is a list, so probably hard to use as the "canonical" name?

ehelms:
> certs::cname is a list, so probably hard to use as the "canonical" name?

Fair point, this is in part why I had originally back in the RFC wondered about a single parameter to set a load balancer with and allow the code to do with it what it needs in the immediate time frame and future changes. Having individual parameters provides more power for configuration but more parameters to manage and coorindate.

ehelms:
This has undergone a bit of re-think looking to the use of `foreman_proxy::registration_url` to drive the parameter setting. This keeps the underlying changes internal and not user facing and simplifies the code as well. @ekohl per our discussion

manifests/init.pp
```diff
@@ -264,6 +265,12 @@
     $apache_https_key = undef
     $apache_https_ca = undef
     $apache_https_chain = undef
+
+    if $foreman_proxy::registration_url {
+      $client_facing_servername = foreman_proxy_content::host_from_url($foreman_proxy::registration_url)
```
ekohl, line 270: I was wondering if we should add a failsafe
```suggestion
      $client_facing_servername = foreman_proxy_content::host_from_url($foreman_proxy::registration_url)
      unless $client_facing_servername in $cname {
        fail('...')
      }
```

But then I realized we don't pass `serveraliases` to `pulpcore` at all. In https://github.com/theforeman/puppet-foreman_proxy_content/pull/435/files#r1010674622 I did that and https://github.com/theforeman/puppet-pulpcore/pull/269 has been merged now.

Thoughts on this?

manifests/init.pp
```diff
@@ -357,20 +364,19 @@
   }
   include pulpcore::plugin::certguard # Required to be present by Katello when syncing a content proxy
 
-  $rhsm_url = "https://${servername}:${rhsm_port}${rhsm_path}"
+  $rhsm_url = "https://${client_facing_servername}:${rhsm_port}${rhsm_path}"
 
   class { 'foreman_proxy::plugin::pulp':
     pulpcore_enabled      => true,
     pulpcore_mirror       => $pulpcore_mirror,
-    pulpcore_api_url      => "https://${servername}",
```
ekohl, line 365: We should continue passing in the API URL, just like it was prior to this patch.

ekohl (changes requested):

manifests/init.pp
ehelms: I like the idea behind the check

manifests/init.pp
```diff
@@ -264,6 +265,16 @@
     $apache_https_key = undef
     $apache_https_ca = undef
     $apache_https_chain = undef
+
+    if $foreman_proxy::registration_url {
+      $client_facing_servername = foreman_proxy_content::host_from_url($foreman_proxy::registration_url)
+
+      unless $client_facing_servername in $certs::cname {
```
ekohl, line 272: Can we add `$serveraliases = $certs::apache::cname` around line 260 and an equivalent around line 247 (there you can probably use `undef` since it doesn't matter, or `$foreman::config::apache::serveraliases` if you want to have the proper value).

Then here you can use:
```suggestion
      unless $client_facing_servername in $serveraliases {
```

I was hoping to pass `serveraliases` to the `pulpcore` class around line 295, but see I'd need to release puppet-pulpcore for that. Still, I wouldn't mind doing that so we have a correct setup where Apache is aware of the server names it should accept.

ehelms: Sure thing! I've added that specification

ekohl: Looks like I opened https://github.com/theforeman/puppet-pulpcore/pull/272 in early November, but it was never reviewed. Just updated now. Please make sure you set that as a lower bound in metadata.json as well.

ehelms: With my current diff on this PR, I am not seeing how it would put a requirement on https://github.com/theforeman/puppet-pulpcore/pull/272 ? That feels like a separate change.

ekohl: That's this part that I mentioned:
> I was hoping to pass serveraliases to the pulpcore class around line 295, but see I'd need to release puppet-pulpcore for that. Still, I wouldn't mind doing that so we have a correct setup where Apache is aware of the server names it should accept.

You're right you didn't implement that, but I think we should.

ehelms: I added this change as a additional commit, while it requires the previous work it's not required to tackle the original commit. The second commit contains the `metadata.json` update as well.

ekohl (approved):
Thanks! I appreciate this.

PR #435, opened 2022-11-01
Fixes #35681 - Expose hostname and cname parameters

ekohl:
Alternative to https://github.com/theforeman/puppet-foreman_proxy_content/pull/432. Passing the aliases to Apache is especially important when there are multiple vhosts or when using [mod_md](https://httpd.apache.org/docs/2.4/mod/mod_md.html). While today we don't do either, having the option is good. It can also help with debugging since sosreport does collect the vhosts, but not the certificates. If ServerAliases matches what's on the certificate, that can give a hint.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -22,6 +22,12 @@
 #
 # $pulpcore_mirror::                           Deploy Pulp to be used as a mirror
 #
+# $hostname::                                  Set the hostname where clients connect to.
+#                                              It ends up on the certificate as the Common Name and in Apache's ServerName.
+#
+# $cname::                                     Set the alternative names where clients can connect to.
+#                                              It ends up on the certificate as CNAMEs and in Apache's ServerAliases.
+#
 # === Advanced parameters:
 #
 # $puppet::                                    Enable puppet
@@ -148,6 +154,9 @@ class foreman_proxy_content (
   Variant[Stdlib::Absolutepath, Array[Stdlib::Absolutepath]] $pulpcore_additional_import_paths = [],
   Variant[Stdlib::Absolutepath, Array[Stdlib::Absolutepath]] $pulpcore_additional_export_paths = [],
   Boolean $pulpcore_telemetry = false,
+
+  Optional[Stdlib::Fqdn] $hostname = undef,
+  Array[Stdlib::Fqdn] $cname = [],
 ) inherits foreman_proxy_content::params {
   include certs
   include foreman_proxy
@@ -169,6 +178,13 @@ class foreman_proxy_content (
 
   include certs::foreman_proxy
   Class['certs::foreman_proxy'] ~> Service['foreman-proxy']
+
+  if $reverse_proxy_real or !$shared_with_foreman_vhost {
+    class { 'certs::apache':
+      hostname => $hostname,
+      cname    => $cname,
+    }
+  }
 
   if $reverse_proxy_real {
     foreman_proxy_content::reverse_proxy { "rhsm-pulpcore-https-${reverse_proxy_port}":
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -279,6 +292,7 @@
     apache_https_chain             => $apache_https_chain,
     apache_vhost_priority          => $priority,
     servername                     => $servername,
+    serveraliases                  => $cname,
```
ekohl, line 295: Relies on https://github.com/theforeman/puppet-pulpcore/pull/269.

ekohl:
I just noticed I forgot to add a crucial bit: the actual 2 class parameters. Perhaps this confused you @ehelms?

manifests/init.pp
```diff
@@ -170,6 +179,13 @@
   include certs::foreman_proxy
   Class['certs::foreman_proxy'] ~> Service['foreman-proxy']
 
+  if $reverse_proxy_real or !$shared_with_foreman_vhost {
+    class { 'certs::apache':
```
ekohl, line 183: I think this doesn't result in a duplicate class declaration with katello (since it should be scoped to just content proxies), but that'd be good to check.

manifests/init.pp
```diff
@@ -148,6 +154,9 @@
   Variant[Stdlib::Absolutepath, Array[Stdlib::Absolutepath]] $pulpcore_additional_import_paths = [],
   Variant[Stdlib::Absolutepath, Array[Stdlib::Absolutepath]] $pulpcore_additional_export_paths = [],
   Boolean $pulpcore_telemetry = false,
+
+  Optional[Stdlib::Fqdn] $hostname = undef,
```
ekohl, line 158: Making it undef by default means that if the FQDN changes, the user gets new certificates. This has implications on the hostname rename script.

ehelms:
As of today, users are instructed to install a load balanced content proxy by specifying `--certs-cname "loadbalancer.example.com" \` as part of the installation parameters. With this change, would the expectation be that that is dropped and replaced by `--foreman-proxy-content-cname loadbalancer.example.com` ?

ekohl:
I'd actually say `--foreman-proxy-content-hostname loadbalancer.example.com --foreman-proxy-content-cname $HOSTNAME` so that it properly identifies primarily as the service name and secondarily as the server itself.

ekohl:
I'm mostly worried about the name on the client certificate. There it should be the host FQDN (unless it's truly load balanced and Foreman doesn't know about each single Smart Proxy, but we're not there).

Did I mention Satellite should have never called this a supported feature?

ehelms:
> I'm mostly worried about the name on the client certificate

Just to be clear, which client certificate?

ekohl:
The ones Foreman Proxy presents to Foreman. Though that does prefer the SAN over the CN and both the hostname and cname(s) are set as SAN entries. Foreman will probably ignore the loadbalancer hostname since there's no matching Smart Proxy with that as the URL.

ehelms:
I still don't think I fully grasp why to prefer the loadbalancer as the CN rather than as a cname. As the load balancer is kinda like a secondary capability, would we not want to then prefer the CN as the FQDN? I admit I don't know what the state of the art is :smile:

ehelms:
> And AFAIK clients do need to talk to the Pulp API for some things.

This is not true for yum. Are there some content types where this is true?

ekohl:
I think it was for containers, but don't exactly recall.

ehelms:
> And AFAIK clients do need to talk to the Pulp API for some things.

Did a little asking and Ansible, Container, and Python plugins all requires API access.

ekohl:
@ehelms reviving this thread. Did you ever find out if client requests are read-only for those plugins?

ehelms:
> @ehelms reviving this thread. Did you ever find out if client requests are read-only for those plugins?

There is always the case of uploading content, consumption seems to be largely read-only but there is nothing that enforces this across the board for content plugins.

ekohl:
We merged 1e2f8390866abeced23819fdf8aea5a0956b0387 which correctly sets the CNAMEs to match the certificates. This is good enough for now.

PR #440, opened 2023-01-24
Refs #35985 - Rely on theforeman/puppet to do integration

ekohl:
The integration is moved into theforeman/puppet so this is no longer needed.

Depends on https://github.com/theforeman/puppet-puppet/pull/862.


ekohl:
puppet-systemd 4 came out and broke puppet-redis. This isn't a problem in production, since there we're still on 3 but does break our test suite.

ekohl:
I've opened https://github.com/voxpupuli/puppet-systemd/pull/313 which is probably the easiest fix in the short term.

ehelms (approved):
I know there is an order, or group that have to go in together. This one looks good as the stand-alone change, only merge when the rest are ready.

PR #442, opened 2023-04-24
Fixes #36854 - Use HTTP/2 when proxying to Foreman

ekohl:
This aims to use mod_proxy_http2, which utilizes HTTP/2 to a backend server. It depends on a feature released in puppetlabs-apache 10.1.1.

Overall testing shows the connection is much more stable and reliable than with HTTP/1.1, especially in higher concurrency. It is recommended to keep this on, but there is an option to disable it if for some reason it doesn't work.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -26,6 +26,8 @@
 #
 # $reverse_proxy_port::                        Reverse proxy listening port
 #
+# $reverse_proxy_backend_protocol::            Configure the protocol used by the reverse proxy to connect to Foreman
+#
 # $pulpcore_allowed_content_checksums::        List of checksums to use for pulpcore content operations
 #
 # $pulpcore_manage_postgresql::                Manage the Pulpcore PostgreSQL database.
@@ -81,6 +83,7 @@ class foreman_proxy_content (
 
   Boolean $reverse_proxy = false,
   Stdlib::Port $reverse_proxy_port = 8443,
+  Enum['h2', 'https'] $reverse_proxy_backend_protocol = 'h2',
 
   Boolean $enable_yum = true,
   Boolean $enable_file = true,
@@ -120,6 +123,7 @@ class foreman_proxy_content (
   $foreman_url = $foreman_proxy::foreman_base_url
   $foreman_host = foreman_proxy_content::host_from_url($foreman_url)
   $reverse_proxy_real = $pulpcore_mirror or $reverse_proxy
+  $proxy_foreman_url = $foreman_url.regsubst('https://', "${reverse_proxy_backend_protocol}://")
 
   # TODO: make it configurable
   # https://github.com/theforeman/puppet-foreman_proxy_content/issues/407
@@ -137,7 +141,7 @@ class foreman_proxy_content (
 
   if $reverse_proxy_real {
     foreman_proxy_content::reverse_proxy { "rhsm-pulpcore-https-${reverse_proxy_port}":
-      path_url_map => { '/' => "${foreman_url}/" },
+      path_url_map => { '/' => "${proxy_foreman_url}/" },
       port         => $reverse_proxy_port,
       priority     => '10',
       before       => Class['pulpcore::apache'],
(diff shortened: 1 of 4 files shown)


ekohl:
Rebased and now that the feature in puppetlabs-apache is merged I've added the proper metadata.

ehelms:
If I understand correctly, this will mean that incoming HTTP 1.1 and HTTP/2 connections will both be handled. When the proxy communicates with Foreman only HTTP/2 will be used. And since we have already enabled HTTP/2 by default in a prior release we have no worry about HTTP/2 being unavailable on the Apache on the Foreman box itself and thus should have no lack of continuity.

ekohl:
That is indeed also my theory. I've been discussing this with @pablomh who has been testing this. In the light of the HTTP/2 attacks I now wonder if we need a parameter to toggle this if we decide we need to force things back to HTTP/1.1.

ehelms:
> That is indeed also my theory. I've been discussing this with @pablomh who has been testing this. In the light of the HTTP/2 attacks I now wonder if we need a parameter to toggle this if we decide we need to force things back to HTTP/1.1.

While Apache has been denoted as not affected, it couldn't hurt to have our own switch to flip.

ekohl:
Updated with a boolean and a Redmine issue.

manifests/init.pp
```diff
@@ -81,6 +83,7 @@
 
   Boolean $reverse_proxy = false,
   Stdlib::Port $reverse_proxy_port = 8443,
+  Boolean $reverse_proxy_use_http2_to_backend = true,
```
evgeni, line 86: I wonder if
```puppet
Enum['h2','https'] $backend_protocol = 'h2'
```
is cleaner?

ekohl: Not sure about the answer. On the one hand, it feels like an internal implementation detail which protocol we use. On the other hand, I can see some flexibility here. Other options are (in theory) h2c (HTTP/2 cleartext) and possibly in the future HTTP/3.

evgeni: Isn't `reverse_proxy_use_http2_to_backend` leaking the same detail? ;)
But yeah, the flexibility to add H3 in the future was what I was aiming at.

Wouldn't block on this, but wanted to open up the discussion.

ekohl: Updated

evgeni (approved):

pablomh:
We have analyzed this patch and here are the findings:

The test consisted on running 5 batches of incremental concurrent registrations, using a 8 CPU / 32 GiB RAM Satellite server, using `--tuning medium --foreman-db-pool 128` as tuning parameters and with and without the patch.

- Baseline run:

image

- Patched run:

image

- Comparison of average number of failed registration attempts:

image

- Comparison of best results without the patch and worse with the patch:

image

Our interpretation is that it tends to behave much better and show much more consistent numbers (without the patch we see two runs with "good enough" numbers and 3 with bad results), with the last chart showing that even the worse results with the patch applied are in line with the better results without the patch.

ekohl:
That number 150 looked familiar. That happens to be the default that used to be applied for `MaxRequestWorkers` when using the medium tuning. https://github.com/theforeman/foreman-installer/commit/068d1f69c8f6f316aae9953d4e8d74959159efc7 got rid of it and now the default should be higher (400). Specifically note that the graphs show it drops off just below 150, which probably is the number 144 that is mentioned in the warning.

I now wonder if both patches together yield an even better result. This patch on the content proxy and the other on the Foreman server.

ekohl:
Good point, I was confused with another setting. It does make me think though. If we assume it is indeed 144 where it starts to fall off. Then Foreman has configured Puma with 16 threads max. 144/16 is 9, so is the setup using 9 Puma workers by any chance? If so, it would suggest that Apache is giving it the maximum concurrency and the next focus should be on Foreman itself. Increase the number of workers or further optimize the endpoints to return faster.

pablomh:
Puma settings were not modified for this testing, an according to the sosreport I have of that system:
```
	Environment: FOREMAN_PUMA_THREADS_MIN=5
	Environment: FOREMAN_PUMA_THREADS_MAX=5
	Environment: FOREMAN_PUMA_WORKERS=12
```

ekohl:
Then I have a hard time explaining that. With those numbers you'd expect a concurrency of 60. Perhaps I can find time to apply https://docs.theforeman.org/nightly/Monitoring_Project/index-katello.html and test it myself.

PR #455, opened 2023-08-15
Remove katello-agent

wbclark:


ehelms:
Before we merge any of the changes, we will need staged updates for our pipeline tests, https://github.com/theforeman/forklift/blob/master/bats/fb-katello-client.bats#L10 and any other related katello-agent tests will need scoping. The puppet module tests now remove the testing of an enabled qpid being disabled so we should consider either:

 1. Keeping a test here that first enables all the qpid bits and then tests that the module disables it properly
 2. Add a pipeline upgrade test, similar to how we did Pulp 2 removal (https://github.com/theforeman/forklift/blob/master/pipelines/upgrade/07-server_to_final.yml#L24) that ensures katello-agent is enabled before the test upgrades to nightly / 4.10

@evgeni if you have any thoughts on this

evgeni:
I think I'd prefer the forklift/bats route, as that's what users will be facing: having a 4.9 with k-a, upgrading to 4.10 and abandoning k-a.

manifests/init.pp
```diff
@@ -196,36 +159,9 @@
...
-    ensure        => bool2str($enable_katello_agent, 'present', 'absent'),
-    agent_addr    => $qpid_router_agent_addr,
-    agent_port    => $qpid_router_agent_port,
-    ssl_ciphers   => $qpid_router_ssl_ciphers,
-    ssl_protocols => $qpid_router_ssl_protocols,
-    logging_level => $qpid_router_logging_level,
-    logging       => $qpid_router_logging,
-    logging_path  => $qpid_router_logging_path,
-  }
+  class { 'foreman_proxy_content::dispatch_router': }
```
ekohl, line 162: If you already have contain (as below) this line is redundant
```suggestion
```

spec/acceptance/dispatch_router_spec.rb
```diff
@@ -5,28 +5,7 @@
...
-  end
-
-  context 'with ensure absent' do
-    it_behaves_like 'an idempotent resource' do
-      let(:manifest) do
-        <<-PUPPET
-        class { 'foreman_proxy_content::dispatch_router':
-          ensure => 'absent',
-        }
+        class { 'foreman_proxy_content::dispatch_router': }
```
ekohl, line 8: Why change the previous line?
```suggestion
        include foreman_proxy_content::dispatch_router
```

manifests/dispatch_router.pp
```diff
@@ -1,81 +1,8 @@
 # @summary Install and configure Qpid Dispatch Router
```
ekohl, line 1: This is no longer correct.

spec/acceptance/dispatch_router_spec.rb
wbclark: I think it's just a coincidence of the way I made the change in the editor vs. the way the diff got rendered. I changed 'with ensure absent' to 'with default parameters' and the next couple lines were the same.

To your point, I can change this to an include statement since we're no longer passing parameters here.

manifests/dispatch_router.pp
wbclark: I could change this to 'Uninstall Qpid Dispatch Router' but this comment got me thinking that the class could just be removed and what little content remains could be moved into init.pp

wbclark:
Thanks for the feedback @ekohl 

I amended the commit with initial changes based on suggestions

I then added a second commit which further removes the dispatch_router class entirely, moving the little remaining code for cleanup into the base class.

I kept that as a 2nd commit for ease of review, but intend for it to be squashed before merging.

ehelms:
I think this generally looks good, without a test with it setup and then running the module it's harder to tell if we missed something or not. I think in hindsight it would have been simpler to just leave all the code alone but enforce a `false` on the parameter. That would have reduce the variance being introduced and then allowed ripping all the code out in the following release.

@ekohl please also have a look

ekohl (approved):

PR #469, opened 2024-01-10
Support EL9

evgeni:


evgeni:
Whoa, it's green?!

ekohl:
That makes me think we don't have enough test coverage, rather than that it all works ;) But nice!

evgeni:
> That makes me think we don't have enough test coverage, rather than that it all works ;) But nice!

Yeah, it mostly just verifies the app starts up. But hey, it's something.

ehelms (approved):

PR #476, opened 2024-02-21
Remove qpid class

ehelms:


evgeni (approved):

evgeni (changes requested):
fixtures still refer to qpid

PR #479, opened 2024-04-05
Fixes #37325 - make postgres the container gateway default DB

ianballou:
Draft PR to instantiate the postgres database for the container gateway. The main PR is here: https://github.com/theforeman/puppet-foreman_proxy/pull/835

I'm not totally sure if foreman_proxy_content is the proper spot to create the DB, but it seemed better than doing it in foreman_proxy and adding a postgres library requirement there.

I'm looking for general feedback on what things I might be missing between the two PRs since I don't work on Puppet modules often.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -280,6 +280,16 @@ class foreman_proxy_content (
         pulp_endpoint => "https://${servername}",
       }
     }
+    include postgresql::client
+    include postgresql::server
+    include postgresql::server::contrib
+    postgresql::server::db { $foreman_proxy::plugin::postgresql_database:
+      user     => $foreman_proxy::plugin::postgresql_user,
+      password => postgresql::postgresql_password($foreman_proxy::plugin::postgresql_user, $foreman_proxy::plugin::postgresql_db_password),
+      encoding => 'utf8',
+      locale   => 'en_US.utf8',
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -280,6 +280,16 @@
         pulp_endpoint => "https://${servername}",
       }
     }
+    include postgresql::client
+    include postgresql::server
+    include postgresql::server::contrib
```
ekohl, line 285: Do you need extensions from contrib?

manifests/init.pp
```diff
@@ -280,6 +280,16 @@
         pulp_endpoint => "https://${servername}",
       }
     }
+    include postgresql::client
```
ekohl, line 283: Normally `client` is not needed.

ekohl (changes requested):
As mentioned in https://github.com/theforeman/puppet-foreman_proxy/pull/835#pullrequestreview-1983574442, I'd prefer to manage the DB there.

I do wonder about external DB support. AFAIK we don't currently support that on content proxies, but if we would then we'd need to account for it.

ianballou:
Closing to put this work in foreman_proxy. (https://github.com/theforeman/puppet-foreman_proxy/pull/835)

PR #480, opened 2024-05-13
enforce xz compression for consumer rpm

evgeni:
EL9+ defaults to zstd, but EL7 can't consume that


evgeni:
This brings back memories I hoped not to see ever again.

https://github.com/Katello/katello/pull/70

ehelms (approved):

ekohl:
A Redmine issue for this would have been nice because we may want to cherry pick it.

PR #481, opened 2024-05-17
Set Apache parameters in ci to reflect the real deployment

ekohl:
The installer has these values set and this makes the test more representative of the real world.


ehelms (approved):
Once tests are happy

ehelms (approved):

spec/acceptance/hieradata/common.yaml
```diff
@@ -1,5 +1,13 @@
 ---
 apache::default_mods: false
+# This is normally set in the installer, but the bootstrap_rpm test needs it
```
ekohl, line 3: I didn't want to fully dive into it to fix the tests. This is already close enough now.

PR #487, opened 2024-07-30
Fixes #37701: Default 8443 reverse proxy to off

ehelms:
This prepares us to remove this in the next release by getting users used to this being off by default. For upgraded systems, this will disable port 8443 and any user who actually still needs it will have to explicitly enable it.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -11,6 +11,7 @@ fixtures:
       repo: "https://github.com/puppetlabs/puppetlabs-cron_core"
       puppet_version: ">= 6.0.0"
     datacat:       "https://github.com/richardc/puppet-datacat"
+    dns:           "https://github.com/theforeman/puppet-dns.git"
     candlepin:     "https://github.com/theforeman/puppet-candlepin.git"
     foreman_proxy: "https://github.com/theforeman/puppet-foreman_proxy.git"
     foreman:       "https://github.com/theforeman/puppet-foreman"
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -122,7 +122,7 @@ class foreman_proxy_content (
 
   $foreman_url = $foreman_proxy::foreman_base_url
   $foreman_host = foreman_proxy_content::host_from_url($foreman_url)
-  $reverse_proxy_real = $pulpcore_mirror or $reverse_proxy
+  $reverse_proxy_real = $pulpcore_mirror and $reverse_proxy
   $proxy_foreman_url = $foreman_url.regsubst('https://', "${reverse_proxy_backend_protocol}://")
 
   # TODO: make it configurable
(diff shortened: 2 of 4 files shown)


ekohl (changes requested):
I'd like to have a Redmine issue for this because I think it'll also deserve a release note. This is a backwards incompatible change now. It will also mean we disable port 8443 on existing installations. Is there an accompanying foreman-installer change to maintain it on existing installations?

ehelms:
> I'd like to have a Redmine issue for this because I think it'll also deserve a release note. This is a backwards incompatible change now. It will also mean we disable port 8443 on existing installations. Is there an accompanying foreman-installer change to maintain it on existing installations?

This feels borderline backwards incompatible? This can still be turned on, so it is backwards compatible when the option is provided.

The idea is to disable this for everyone, including upgrades. Users who need this should take a direct action to re-enable it so that they understand the choice that they are making. The only things that should be affected by this are:

 * clients that have existed for 3+ years and have not updated the katello-ca-consumer RPM
 * users using the 8443 endpoint for API access (which we are wanting to close off)

ekohl:
> This feels borderline backwards incompatible? This can still be turned on, so it is backwards compatible when the option is provided.

I meant backwards incompatible from a Puppet module point of view.

> The idea is to disable this for everyone, including upgrades. Users who need this should take a direct action to re-enable it so that they understand the choice that they are making. The only things that should be affected by this are:

Sounds fair to me. Still, an upgrade warning is IMHO a good thing for this. Better warned too often than too little.

ehelms:
> > This feels borderline backwards incompatible? This can still be turned on, so it is backwards compatible when the option is provided.
> 
> I meant backwards incompatible from a Puppet module point of view.

We could set `reverse_proxy` to `true` here to avoid this and only set it to `false` in the installer. If that's useful?


> Sounds fair to me. Still, an upgrade warning is IMHO a good thing for this. Better warned too often than too little.

What sort of upgrade warning do you envision?

ekohl (approved):
> We could set `reverse_proxy` to `true` here to avoid this and only set it to `false` in the installer. If that's useful?

The parameter already exists, so on existing installations it will be stored in the answers file already. That won't be helpful.

> What sort of upgrade warning do you envision?

Something in https://github.com/theforeman/foreman-documentation/blob/master/guides/doc-Release_Notes/topics/katello.adoc#upgrade-warnings that tells users that if they upgrade, the reverse proxy on port 8443 will be turned off. Ideally with a link to some procedure to upgrade clients and/or instructions on how to turn it back on.

ehelms:
> Something in https://github.com/theforeman/foreman-documentation/blob/master/guides/doc-Release_Notes/topics/katello.adoc#upgrade-warnings that tells users that if they upgrade, the reverse proxy on port 8443 will be turned off. Ideally with a link to some procedure to upgrade clients and/or instructions on how to turn it back on.

Got a draft up for it -- https://github.com/theforeman/foreman-documentation/pull/3180

nixfu:
BTW, I am not sure this is accurate.   The 443 does not reverse proxy things like the foreman API.   We use the reverse proxy on 8443 today to do foreman api calls, for things like clients checking if they are in the foreman inventory, and inserting themselves if they are not etc..

evgeni:
Correct, that's intentional

PR #490, opened 2024-08-14
Fixes #37620 - Use the correct docroot in the reverse proxy setup

ekohl:
This is a code dump of what I was working on.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -146,22 +146,6 @@ class foreman_proxy_content (
     }
   }
 
-  if $pulpcore_mirror {
-    $pulpcore_https_vhost_name = "rhsm-pulpcore-https-${rhsm_port}"
-
-    if $rhsm_port != $reverse_proxy_port {
-      foreman_proxy_content::reverse_proxy { $pulpcore_https_vhost_name:
-        path_url_map => {
-          $rhsm_path     => "${proxy_foreman_url}${rhsm_path}",
-          $insights_path => "${proxy_foreman_url}${insights_path}",
-        },
-        port         => $rhsm_port,
-        priority     => '10',
-        before       => Class['pulpcore::apache'],
-      }
-    }
-  }
-
   include foreman_proxy_content::pub_dir
 
   if $pulpcore_mirror {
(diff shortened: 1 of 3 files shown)


ehelms:
Do you think there is any way to add an acceptance test for either `/cgi-bin` or the Apache multiviews directory listing?

ekohl:
Looking at what it now serves is an a directory listing without entries:
```console
$ curl https://localhost -k
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<html>
 <head>
  <title>Index of /</title>
 </head>
 <body>
<h1>Index of /</h1>
  <table>
   <tr><th valign="top"><img src="/icons/blank.gif" alt="[ICO]"></th><th><a href="?C=N;O=D">Name</a></th><th><a href="?C=M;O=A">Last modified</a></th><th><a href="?C=S;O=A">Size</a></th><th><a href="?C=D;O=A">Description</a></th></tr>
   <tr><th colspan="5"><hr></th></tr>
   <tr><th colspan="5"><hr></th></tr>
</table>
</body></html>
```

It's slightly misleading that it doesn't list `/pub` even though it will be served. I'm tempted to always add `/var/lib/pulp/pulpcore_static/index.html` which would also avoid stupid scanners thinking multiviews is still a thing. Or should we add some placeholder there to inform the user what it's for?

ehelms:
> It's slightly misleading that it doesn't list `/pub` even though it will be served. I'm tempted to always add `/var/lib/pulp/pulpcore_static/index.html` which would also avoid stupid scanners thinking multiviews is still a thing. Or should we add some placeholder there to inform the user what it's for?

I'm OK with a blank file being added. Would be nice to document why we have it - which could be through a comment or by stating that on the file itself.

ekohl:
Current version instead disables indexes and now it shows a HTTP 403 page. I didn't include a test because we don't have the [curl_command](https://github.com/theforeman/puppet-pulpcore/blob/master/spec/support/acceptance/curl_command.rb) helper here. @evgeni mentioned he'd like it some other place as well so I intend to move it to [voxpupuli-acceptance](https://github.com/voxpupuli/voxpupuli-acceptance) but I'd prefer to not block this PR on it.

ehelms (approved):
I got a 403 on `/` and all other endpoints I should browse are visible.

PR #494, opened 2024-08-21
Drop redundant update-ca-trust enable step

ekohl:
On EL 7 & 8 the argument is ignored, so it effectively ran twice. On EL 9 since ca-certificates-0:2024.2.69_v8.0.303-91.3.el9 it does check the argument and changes behavior. The on 91.3 it hard fails on enable while 91.4 yields a deprecation warning. It still effectively runs twice.

This drops the redundant step because it only needs to extract after the new CA has been added.

See https://github.com/voxpupuli/puppet-trusted_ca/pull/69 for a lot more context on this.

### DIFF
--- a/lib/puppet/provider/rhsm_reconfigure_script/rhsm_reconfigure_script.rb
+++ b/lib/puppet/provider/rhsm_reconfigure_script/rhsm_reconfigure_script.rb
@@ -125,9 +125,8 @@ Puppet::Type.type(:rhsm_reconfigure_script).provide(:rhsm_reconfigure_script) do
 
       # also add the katello ca cert to the system wide ca cert store
       if [ -d $CA_TRUST_ANCHORS ]; then
-        update-ca-trust enable
         cp $CERT_DIR/$KATELLO_SERVER_CA_CERT $CA_TRUST_ANCHORS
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/rhsm_reconfigure_script/rhsm_reconfigure_script.rb
```diff
@@ -125,7 +125,6 @@ def script_template
 
       # also add the katello ca cert to the system wide ca cert store
       if [ -d $CA_TRUST_ANCHORS ]; then
-        update-ca-trust enable
         cp $CERT_DIR/$KATELLO_SERVER_CA_CERT $CA_TRUST_ANCHORS
         update-ca-trust
```
ekohl, line 129: Should this be changed to explicitly extract? Similar to https://github.com/voxpupuli/puppet-trusted_ca/pull/69#discussion_r1724424510
```suggestion
        update-ca-trust extract
```

evgeni: I'd do that, yeah

ehelms (approved):

PR #496, opened 2024-10-14
Drop port 8443 as a reverse proxy

ehelms:


ehelms:
Since we are making a backwards incompatible change, do we want to drop `reverse_proxy_backend_protocol` as a parameter as well? We've been operating with http/2 for a bit now.

ekohl:
I'm not sure I intended that as a temporary parameter. While we don't intend typical users to change this, I'd still like to keep the options open to tune this in some edge cases.

ehelms:
> I'm not sure I intended that as a temporary parameter. While we don't intend typical users to change this, I'd still like to keep the options open to tune this in some edge cases.

Fair enough.

evgeni (approved):

manifests/reverse_proxy.pp
```diff
@@ -19,7 +19,7 @@
 define foreman_proxy_content::reverse_proxy (
```
ekohl, line 19: Do we still need this as a define? Taking a step back, let's look at the whole picture.

This now defines a vhost, but Pulp can also define a vhost. So can Foreman. What this then adds is the RHSM bits, which we [also do in puppet-katello](https://github.com/theforeman/puppet-katello/blob/master/files/katello-apache-ssl.conf).

Perhaps we should capture this as a future refactor, but I feel there's room to improve on alignment here.

On the other hand, they are also conceptually different: with puppet-katello it's actual application code handling things while with puppet-foreman_proxy_content it's forwarded to the real application.

ehelms: I'll take a note to have a look at re-factoring this week. That's always a good idea. I think the right answer is to do that after we've done the minimal change to extract this functionality first.

ekohl (approved):
I think this is OK now, but if we refactor it to a class we can actually use Hiera to tune things (like `proxy_pass_params`). Since this is backwards incompatible, I think it's a good chance.

PR #497, opened 2024-11-04
Mark compatible with theforeman/pulpcore 11.x

ekohl:


bastelfreak (approved):
I adjusted the title. But feel free to change it if you prefer the old name.

ekohl:
Oh you're absolutely right that I intended to include a version.

PR #499, opened 2024-11-04
Release 30.0.0

ekohl:


CHANGELOG.md
```diff
@@ -1,8 +1,21 @@
 # Changelog
 
-## [29.0.0](https://github.com/theforeman/puppet-foreman_proxy_content/tree/29.0.0) (2024-08-21)
```
ekohl, line 3: This is very confusing, so I'll just point out that with 29.0.1 I failed to bump it before generating and turns out the changelog generator is just fine with having the same release version twice in the changelog.

ehelms (approved):

PR #501, opened 2024-12-20
Add enablement parameter for bootstrap RPMs

ehelms:
This will allow us to disable the bootstrap RPM and the consumer RPM when we are ready. The only odd situation is if someone were to re-enable it the RPM counter would start back at 1 which clients with the RPM already would not upgrade to without being forced.

### DIFF
--- a/lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
+++ b/lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
@@ -22,12 +22,28 @@ Puppet::Type.type(:bootstrap_rpm).provide(:bootstrap_rpm) do
     @base_dir = nil
   end
 
-  def exists?
-    copy_sources
-    write_specfile
-    build_rpm
+  def destroy
+    FileUtils.rm(symlink) unless symlink.nil?
 
-    !rpm_changed?
+    all_rpms.each do |rpm|
+      FileUtils.rm(rpm)
+    end
+
+    all_rpms(source: true).each do |rpm|
+      FileUtils.rm(rpm)
+    end
+  end
+
+  def exists?
+    if resource[:ensure] == :absent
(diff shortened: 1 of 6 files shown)


lib/puppet/provider/bootstrap_rpm/bootstrap_rpm.rb
```diff
@@ -22,12 +22,28 @@ def create
...
+    all_rpms(source: true).each do |rpm|
+      FileUtils.rm(rpm)
+    end
+  end
+
   def exists?
-    copy_sources
-    write_specfile
-    build_rpm
+    if resource[:ensure] == :absent
```
ekohl, line 38: This feels like an abuse of `exists`. I suppose it already was and I think this currently doesn't support noop mode, but this really is making it much worse. `exists?` should really answer the question if anything exists.

Looking at the source, I think this is the relevant bit:
https://github.com/puppetlabs/puppet/blob/e227c27540975c25aa22d533a52424a9d2fc886a/lib/puppet/property/ensure.rb#L68-L95

So if anything, it should be changed to:
```ruby
if resource[:ensure] == :absent
  File.exist?(resource[:symlink]) || !all_rpms.empty? || !all_rpms(source: true).empty?
else
  !File.exist?(resource[:symlink]) || all_rpms.empty? || all_rpms(source: true).empty? || rpm_changed?
end
```

Then it will end up calling `.create`:
https://github.com/puppetlabs/puppet/blob/e227c27540975c25aa22d533a52424a9d2fc886a/lib/puppet/property/ensure.rb#L16-L23
Or `.destroy`:
https://github.com/puppetlabs/puppet/blob/e227c27540975c25aa22d533a52424a9d2fc886a/lib/puppet/property/ensure.rb#L25-L32

It's still not very nice, but since it's going away I think that's ok.

PR #503, opened 2025-01-14
Drop EL8 support -- Foreman/Katello do not support EL8 anymore

evgeni:


ekohl (approved):

ekohl (changes requested):
Wait, let's also drop https://github.com/theforeman/puppet-foreman_proxy_content/blob/e24f94194ebd7858f4ce9c34e5a671a30009873c/spec/setup_acceptance_node.pp#L22-L26

evgeni:
Done.

ekohl (approved):
Now it died a little more ;)

PR #514, opened 2025-03-05
Use the Apache CA cert as SSL proxy cert

ekohl:
This replaces the `certs::ca_cert` variable with `certs::apache::apache_ca_cert` because the former has been removed in git. It's safe to assume Foreman is using the same CA certificate as the vhost itself is using.


ehelms:
See https://github.com/theforeman/puppet-foreman_proxy_content/pull/515 for what I think is a more correct method

ehelms:
Closed in favor of https://github.com/theforeman/puppet-foreman_proxy_content/pull/515

PR #515, opened 2025-03-05
Use correct CA variables in reverse proxy

ehelms:
The removal of parameters in puppet-certs that were not used internally were used externally -- this updates to use current canonical parameters.


manifests/reverse_proxy.pp
```diff
@@ -54,12 +54,12 @@
     ssl_options            => ['+StdEnvVars', '+ExportCertData', '+FakeBasicAuth'],
     ssl                    => true,
     ssl_proxyengine        => true,
-    ssl_proxy_ca_cert      => $certs::ca_cert,
+    ssl_proxy_ca_cert      => $certs::katello_default_ca_cert,
     ssl_proxy_machine_cert => $certs::foreman_proxy::foreman_proxy_ssl_client_bundle,
     ssl_cert               => $certs::apache::apache_cert,
     ssl_key                => $certs::apache::apache_key,
-    ssl_chain              => $certs::apache::apache_ca_cert,
```
ekohl, line 61: I kind of liked the consistency here to rely on https://github.com/theforeman/puppet-certs/blob/9585baa00b9158f8da16c36b76841df10b92fb97/manifests/apache.pp#L70

manifests/reverse_proxy.pp
```diff
@@ -54,12 +54,12 @@
     ssl_options            => ['+StdEnvVars', '+ExportCertData', '+FakeBasicAuth'],
     ssl                    => true,
     ssl_proxyengine        => true,
-    ssl_proxy_ca_cert      => $certs::ca_cert,
+    ssl_proxy_ca_cert      => $certs::katello_default_ca_cert,
```
ekohl, line 57: Isn't this the certificate we expect on the remote server (so Foreman)? https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslproxycacertificatefile states:
> This directive sets the all-in-one file where you can assemble the Certificates of Certification Authorities (CA) whose remote servers you deal with. These are used for Remote Server Authentication. Such a file is simply the concatenation of the various PEM-encoded Certificate files, in order of preference. This can be used alternatively and/or additionally to [SSLProxyCACertificatePath](https://httpd.apache.org/docs/current/mod/mod_ssl.html#sslproxycacertificatepath).

evgeni: Looking at https://github.com/theforeman/puppet-certs/commit/9585baa00b9158f8da16c36b76841df10b92fb97:
```
$ca_cert = "${pki_dir}/certs/${default_ca_name}.crt"
$katello_default_ca_cert = "${pki_dir}/certs/${default_ca_name}.crt"
```

so this is an equivalent change

manifests/reverse_proxy.pp
ehelms: I don't disagree. I think that should come with or after this change https://github.com/theforeman/puppet-certs/pull/492 that moves deployment of it into Apache's class.

manifests/reverse_proxy.pp
ehelms: For this change,to fix tests I think we keep it equivalent. As for your point, this change dates back a long, long time to https://github.com/theforeman/puppet-foreman_proxy_content/pull/54.

The configuration is not setting SSLProxyVerify explicitly, where as we do set SSLClientVerify (granted to optional) so I am guessing this setting doesn't actually matter or do anything at the moment.

manifests/reverse_proxy.pp
ehelms: How about we adopt this change -- https://github.com/theforeman/puppet-certs/pull/497 and then I update this to rely on those two from the Apache class?

ekohl: I like that.

ehelms:
Updated.

manifests/reverse_proxy.pp
```diff
@@ -54,12 +54,12 @@
     ssl_options            => ['+StdEnvVars', '+ExportCertData', '+FakeBasicAuth'],
     ssl                    => true,
     ssl_proxyengine        => true,
-    ssl_proxy_ca_cert      => $certs::ca_cert,
+    ssl_proxy_ca_cert      => $certs::apache::apache_client_ca_cert,
```
ekohl, line 57: Now I'm debating if this should somehow be referring to `$certs::foreman_proxy` because Foreman Proxy is also configured with a CA to trust. This would closely relate to `ssl_proxy_machine_cert`. Have you had any thoughts in that direction?

ekohl (approved):

manifests/reverse_proxy.pp
```diff
@@ -54,12 +54,12 @@
...
     ssl_proxy_machine_cert => $certs::foreman_proxy::foreman_proxy_ssl_client_bundle,
     ssl_cert               => $certs::apache::apache_cert,
     ssl_key                => $certs::apache::apache_key,
-    ssl_chain              => $certs::apache::apache_ca_cert,
-    ssl_ca                 => $certs::apache::ca_cert,
+    ssl_chain              => $certs::apache::apache_client_ca_cert,
+    ssl_ca                 => $certs::apache::apache_ca_cert,
```
ekohl, line 62: Nit: mind swapping the order of this back so the diff is smaller?
```suggestion
    ssl_ca                 => $certs::apache::apache_ca_cert,
    ssl_chain              => $certs::apache::apache_client_ca_cert,
```

ekohl (changes requested):

ekohl (approved):

PR #516, opened 2025-03-11
Do not require File resource for SSL cert in bootstrap RPM class

ehelms:
As the certs are now inputs, we can not guarantee that they will have File resources to require.

This failure is also seen in our nightly pipes:

```

[2025-03-10T22:48:32.995Z]     2025-03-10 22:48:32 [ERROR ] [configure] Could not find resource 'File[/root/ssl-build/katello-server-ca.crt]' in parameter 'require' (file: /usr/share/foreman-installer/modules/foreman_proxy_content/manifests/bootstrap_rpm.pp, line: 23) on node pipe-katello-proxy-nightly-almalinux9.example.com

[2025-03-10T22:48:32.995Z]     2025-03-10 22:48:32 [NOTICE] [configure] System configuration has finished.
```


manifests/bootstrap_rpm.pp
```diff
@@ -19,8 +19,7 @@
 
   include trusted_ca
   trusted_ca::ca { 'katello_server-host-cert':
```
ekohl, line 21: I think it's possible to still have them with a resource collector. From https://www.puppet.com/docs/puppet/8/lang_collectors.html
```puppet
  File <| title == $server_ca_cert |> ->
  trusted_ca::ca { 'katello_server-host-cert':
```

ehelms:
Is that like "if this resource exists create the relationship, otherwise do
nothing"?

manifests/bootstrap_rpm.pp
ehelms: The way puppet-certs is designed and works, the current setup will never have a File resource. When the tarfile is extracted on an external foreman-proxy it just lays it out on disk and skips generate. On Foreman, the only place generate is true, the Puppet custom resources that are used do not auto-generate a File resource. So there is currently no setup where a File resource for these will exist.

ekohl:
Pretty much yes.

manifests/bootstrap_rpm.pp
ekohl: Can we then at least change `class { 'foreman_proxy_content::bootstrap_rpm':` to add `require => Class['certs']` so we know the ordering is correct?

PR #517, opened 2025-04-03
Set client_endpoint on container_gateway plugin

ehelms:
Requires https://github.com/theforeman/puppet-foreman_proxy/pull/863


ianballou (approved):
I wasn't able to do a full test on my proxy since the CNAME wasn't set up right for the installer, but looking at the use of `client_facing_servername`, this appears correct to me.

evgeni (approved):

PR #518, opened 2025-04-04
Add proxy pass entry for lightspeed API

ehelms:
The change [in foreman_rh_cloud](https://github.com/theforeman/foreman_rh_cloud/pull/972) introduces proxying the lightspeed endpoint to the Red Hat cloud. This enables hosts connected through a content proxy to also send their traffic at this new API.


chris1984 (approved):
Tested and works fine

